PR #250, opened 2015-02-09
A couple of small fixes

phobos2077:
- changed <string> to <string.h> header in Game.cpp
- added exception when Texture could not be loaded

### DIFF
--- a/src/Game/Game.cpp
+++ b/src/Game/Game.cpp
@@ -19,7 +19,7 @@
 
 // C++ standard includes
 #include <sstream>
-#include <string>
+#include <string.h>
 
 // Falltergeist includes
 #include "../Audio/AudioMixer.h"
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ falltergeist.cbp
 .idea/
 *.autosave
 /src/nbproject/private/
+nbproject/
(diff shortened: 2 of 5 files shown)


src/Game/Game.cpp
```diff
@@ -19,7 +19,7 @@
 
 // C++ standard includes
 #include <sstream>
-#include <string>
+#include <string.h>
```
alexeevdv: Hm....why <string.h> ?

isage: For strdup.
As for <string> - it's already included in Game.h

alexeevdv: Do we really need strdup here?

isage: Yeah, that's the way putenv works.
>The string pointed to by string becomes part of the environment, so altering the string changes the environment.
But, i'm not sure we need this whole line.

alexeevdv: http://www.qnx.com/developers/docs/6.5.0_sp1/index.jsp?topic=%2Fcom.qnx.doc.neutrino_lib_ref%2Fp%2Fputenv.html
It is needed only when you are not using "literal string". But.... our string is hardcoded and it is quite "literal" :)

isage: Yeah, and will be deallocated when function ends ;)

alexeevdv: There was SDL_putenv function in SDL1.2 - can't find similar function in SDL2.0

isage: There should be SDL_setenv (which actually just calls setenv), but judging by setenv manpage it'll work without duplicating string. So yeah, we could use that (SDL_setenv version, as it deals with platform-specific env) and drop strdup.

Issue #264, opened 2015-02-12
Capture mouse in windowed mode


phobos2077:
When game window is in focus, mouse should be captured (when you move it out of the window, it should stay inside). This will make scrolling by mouse usable in windowed mode.

phobos2077:
Works perfectly, though for some reason I can't use window "close" button anymore, it just captures the mouse and don't close the app. Not a big deal I guess.

PR #266, opened 2015-02-12
Fix for #264

alexeevdv:


phobos2077:
Works perfectly, though for some reason I can't use window "close" button anymore, it just captures the mouse and don't close the app. Not a big deal I guess.

alexeevdv:
Actually we can add shortcut to lock/unlock mouse in window

Issue #255, opened 2015-02-10
UI sound are played with delay


phobos2077:
UI sounds when pressing and releasing buttons are played with very noticeable delay.

tested on Win7 Ultimate x64, integrated audio, standard drivers

PR #268, opened 2015-02-12
fix for #255 by introducing "buffer_size" setting

phobos2077:
Default value 4096 is way too much for 22050 frequency, this resulted in very high latency. I added "buffer_size" setting with default value of 512 (tested with 256 on my machine - seem to work fine, so I doubled this value to be safe).

Testing on my machine revealed that:
* 128 and below starts to crackle and skip sound
* values above 1024 start to give noticeable delay


alexeevdv:
:+1:  Sounds good. Also i checheked out sound in movies. Looks pretty good too

isage:
:+1: 
Maybe we'll need to increase default value on mobile platforms, though

Issue #254, opened 2015-02-10
Popups centered incorrectly on main interface


phobos2077:
Popups should be centered within the active area, and not the whole screen.

Explanation:
![Explanation](http://content.screencast.com/users/Kolosovsky/folders/Jing/media/bc1a3d4d-1a5d-4d3e-9a69-d36d0e2bf9b5/2015-02-10_1917.png)

phobos2077:
This applies to all windows that don't take whole screen. Inventory, exit confirmation, etc.

PR #281, opened 2015-02-12
fixed GameMenu coordinates

vvbalashoff:


vvbalashoff:
closes #254

phobos2077:
Do you think it's OK to use hard-coded dimensions? What if panel size will change?

alexeevdv:
:+1:

isage:
+1 to using Game::getInstance()->locationState()->playerPanelState()->height()

PR #287, opened 2015-02-15
Rework interaction with object using cursor and CursorDropdown

phobos2077:
1) Implemented actions when hovering over objects (with delay) to match original game.
2) Dropdown is opened only after delay and various other tweaks for dropdown.
3) Added sound FX for dropdown.
4) Implemented look_at_p_proc() and description_p_proc() actions for all objects.
5) Added stub method Location::displayMessage() - currently plays monitor.acm and prints message to log
6) Disabled capturing mouse in SDL window because currently it breaks mouse warping which is required for dropdown to work.
7) display_msg() opcode will use Location::displayMessage() method.
8) Fixed invalid number of stack arguments in tile_in_tile_rect() handler.
9) Stubs for all "meta" types in metarule3() handler.

### DIFF
--- a/src/Game/DudeObject.cpp
+++ b/src/Game/DudeObject.cpp
@@ -284,7 +284,7 @@ void GameDudeObject::_generateUi()
 
     if (_ui)
     {
-        _ui->addEventHandler("mouseleftdown", std::bind(&State::Location::onMouseDown, Game::getInstance()->locationState(), std::placeholders::_1, static_cast<GameObject*>(this)));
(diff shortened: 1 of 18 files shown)


alexeevdv:
:+1:

isage:
:+1: 
But, are you sure hotkeys for save/load should be in player panel?

phobos2077:
I did it because save/load buttons are in Menu dialog, and Menu is opened from player panel :)

Issue #288, opened 2015-02-15
Add ability to select which subsystems to monitor in logs


phobos2077:
Some kind of ini setting to list which subsystems (like renderer, scripts, resource manager, etc.) to monitor in logs. This will help to debug only specific systems in real time.

Sh1ft0x0EF:
Hey guys
I've partially implemented the 288 as you can see above but that codebase is pretty outdated already
I think I would sync with current master and reapply it above and then send it as a PR
Any additions/requests?

PR #295, opened 2015-02-17
Opcodes update

phobos2077:
First pass: 
* Updated names for core opcodes.
* Add some comments for missing features.
* Fixed invalid op_dub behavior.
* Implemented op_lookup_string_proc, op_negate, op_play_sfx.
* Added handlers for op_floor, op_critter_set_flee_state.
* Added division by zero check.
* Added missing cases for op_get_critter_stat (along with new radiation property for all Critters).

### DIFF
--- a/src/VM/Handlers/Opcode8044Handler.cpp
+++ b/src/VM/Handlers/Opcode8044Handler.cpp
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2012-2015 Falltergeist Developers.
+ *
+ * This file is part of Falltergeist.
+ *
+ * Falltergeist is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Falltergeist is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Falltergeist.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// C++ standard includes
+
+// Falltergeist includes
+#include "../../Logger.h"
+#include "../../VM/Handlers/Opcode8044Handler.h"
+#include "../../VM/VM.h"
+#include "../../VM/VMStackValue.h"
+#include "../../Exception.h"
+
+// Third party includes
+
+namespace Falltergeist
+{
+
+Opcode8044Handler::Opcode8044Handler(VM* vm) : OpcodeHandler(vm)
+{
+}
+
+void Opcode8044Handler::_run()
+{
+    Logger::debug("SCRIPT") << "[8044] [*] op_floor" << std::endl;
+    auto value = _vm->dataStack()->pop();
+    auto intValue = dynamic_cast<VMStackIntValue*>(value);
+    auto floatValue = dynamic_cast<VMStackFloatValue*>(value);
+    int result = 0;
+    if (intValue) 
+    {
+        result = intValue->value();
+    }
+    else if (floatValue)
+    {
+        result = (int)floatValue->value(); // this is how "floor" originally worked..
+    }
+    _vm->pushDataInteger(result);
+}
+
+}
--- a/src/VM/Handlers/Opcode8044Handler.h
+++ b/src/VM/Handlers/Opcode8044Handler.h
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2012-2015 Falltergeist Developers.
+ *
+ * This file is part of Falltergeist.
+ *
+ * Falltergeist is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Falltergeist is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Falltergeist.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef FALLTERGEIST_OPCODE8044HANDLER_H
+#define FALLTERGEIST_OPCODE8044HANDLER_H
+
+// C++ standard includes
+
+// Falltergeist includes
+#include "../../VM/OpcodeHandler.h"
+
+// Third party includes
+
+namespace Falltergeist
+{
+
+class Opcode8044Handler : public OpcodeHandler
+{
+public:
+    Opcode8044Handler(VM* vm);
+    virtual void _run();
+};
+
+}
+#endif // FALLTERGEIST_OPCODE8045HANDLER_H
--- a/src/VM/Handlers/Opcode8046Handler.cpp
+++ b/src/VM/Handlers/Opcode8046Handler.cpp
@@ -23,6 +23,8 @@
 #include "../../Logger.h"
 #include "../../VM/Handlers/Opcode8046Handler.h"
 #include "../../VM/VM.h"
+#include "VMStackIntValue.h"
+#include "VMStackFloatValue.h"
 
 // Third party includes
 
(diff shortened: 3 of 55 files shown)


src/VM/Handlers/Opcode8044Handler.h
```diff
@@ -0,0 +1,41 @@
...
+
+class Opcode8044Handler : public OpcodeHandler
+{
+public:
+    Opcode8044Handler(VM* vm);
+    virtual void _run();
+};
+
+}
+#endif // FALLTERGEIST_OPCODE8045HANDLER_H
```
alexeevdv: Copy&paste error :)

src/VM/Handlers/Opcode8152Handler.h
```diff
@@ -0,0 +1,42 @@
...
+class Opcode8152Handler : public OpcodeHandler
+{
+public:
+    Opcode8152Handler(VM* vm);
+private:
+    void _run();
+};
+
+}
+#endif
```
alexeevdv: #endif // FALLTERGEIST_OPCODE8152HANDLER_H

src/VM/OpcodeFactory.cpp
```diff
@@ -202,6 +204,8 @@ OpcodeHandler* OpcodeFactory::createOpcode(unsigned int number, VM* vm)
 {
     switch (number)
     {
+        case 0x8000: // O_NOOP
```
alexeevdv: OpcodeFactory *must* return opcode or  nullptr

alexeevdv: Also... i think we can remove all break; statements from factory. 
```       ...
       case 0x8044: return new Opcode8044Handler(vm);
       case 0x8045: return new Opcode8045Handler(vm);
       ...
```

src/VM/VM.cpp
```diff
@@ -171,6 +171,29 @@ void VM::pushDataPointer(void* value, unsigned int type)
...
+            return *value;
+        }
+        throw Exception("VM::popDataString() - stack value is not a string");
+    }
+    throw Exception("VM::popDataString() - stack value is not a pointer");
+}
+
+void VM::pushDataString(std::string value)
+{
+    auto pointer = new VMStackPointerValue(new std::string(value));
```
alexeevdv: this can cause memory leaks. **new** operator without **delete**. I think we cant fix this with VMStackStringValue which will contain std::string

alexeevdv:
Excellent job! Do some fixes and we can merge

src/VM/VM.cpp
```diff
@@ -171,6 +171,29 @@ void VM::pushDataPointer(void* value, unsigned int type)
     _dataStack.push(pointer);
 }
 
+std::string VM::popDataString()
+{
+    auto stackValue = _dataStack.pop();
+    if (stackValue->type() == VMStackValue::TYPE_POINTER)
+    {
+        auto stackPointerValue = dynamic_cast<VMStackPointerValue*>(stackValue);
```
janisozaur: Above you check for `VMStackValue::TYPE_POINTER` and here you use `dynamic_cast`, but few lines below, in the same kind of situation, you use `static_cast`. If there is any danger the cast in line 179 could fail, can you add check against nullptr or otherwise change it to `static_cast`?

src/VM/VM.cpp
```diff
@@ -171,6 +171,29 @@ void VM::pushDataPointer(void* value, unsigned int type)
...
+        {
+            auto value = static_cast<std::string*>(stackPointerValue->value());
+            return *value;
+        }
+        throw Exception("VM::popDataString() - stack value is not a string");
+    }
+    throw Exception("VM::popDataString() - stack value is not a pointer");
+}
+
+void VM::pushDataString(std::string value)
```
janisozaur: Why pass `value` by value?

src/VM/Handlers/Opcode8046Handler.cpp
```diff
@@ -35,9 +37,21 @@ Opcode8046Handler::Opcode8046Handler(VM* vm) : OpcodeHandler(vm)
...
+    {
+        _vm->pushDataInteger(- (dynamic_cast<VMStackIntValue*>(value))->value());
+    }
+    else
+    if (value->type() == VMStackValue::TYPE_FLOAT) 
+    {
+        _vm->pushDataFloat(- (dynamic_cast<VMStackFloatValue*>(value))->value());
+    }
+    else
+    {
```
janisozaur: Log invalid type, possibly with actual type (see https://github.com/falltergeist/falltergeist/blob/master/src/VM/Handlers/Opcode810DHandler.cpp#L52)

src/VM/Handlers/Opcode8046Handler.cpp
```diff
@@ -35,9 +37,21 @@ Opcode8046Handler::Opcode8046Handler(VM* vm) : OpcodeHandler(vm)
...
 void Opcode8046Handler::_run()
 {
-    Logger::debug("SCRIPT") << "[8046] [*] - value (change sign)" << std::endl;
-    auto value = _vm->popDataInteger();
-    _vm-> pushDataInteger(-value);
+    Logger::debug("SCRIPT") << "[8046] [*] op_negate" << std::endl;
+    auto value = _vm->dataStack()->pop();
+    if (value->type() == VMStackValue::TYPE_INTEGER) 
+    {
+        _vm->pushDataInteger(- (dynamic_cast<VMStackIntValue*>(value))->value());
```
janisozaur: `static_cast`, as the type is already known?

src/VM/Handlers/Opcode8044Handler.cpp
```diff
@@ -0,0 +1,56 @@
...
+Opcode8044Handler::Opcode8044Handler(VM* vm) : OpcodeHandler(vm)
+{
+}
+
+void Opcode8044Handler::_run()
+{
+    Logger::debug("SCRIPT") << "[8044] [*] op_floor" << std::endl;
+    auto value = _vm->dataStack()->pop();
+    auto intValue = dynamic_cast<VMStackIntValue*>(value);
+    auto floatValue = dynamic_cast<VMStackFloatValue*>(value);
```
janisozaur: I'd expect scripts to mostly know what they're doing and requesting floor for `float`s more frequently than `int`s. In such case, you could optimize your code a little by first casting to `VMStackFloatValue*` and have an early-out if it was correct. Only cast to `int` if `float` was an incorrect guess.

src/VM/VM.cpp
phobos2077: There was already memory leeks in previous implementation (when string pointer values are created). I was planning to add VMStackStringValue in separate pull request.

src/VM/VM.cpp
phobos2077: I used static_cast for std::string* merely because you can't use dynamic_cast if source type is not a pointer-to-class (void* in this case).

src/VM/VM.cpp
phobos2077: How do you propose?

src/VM/Handlers/Opcode8046Handler.cpp
phobos2077: I don't think it's necessary in this case, because we don't really need specific type here.

src/VM/VM.cpp
janisozaur: by const-ref: `const std::string &value`, this gives compiler hints you're not going to alter the memory in any way and it can use existing object without copy or move

src/VM/Handlers/Opcode8046Handler.cpp
janisozaur: It could indicate bug in the script

phobos2077: I think we need additional method to print run-time warnings for scripts, such as this. So basically there will be two types of errors in scripts:
1) Critical error (like non-object passed where object is expected), prints to debug.log, halts current script invocation, doesn't crash the game.
2) Non-critical error, just print to debug.log. But I don't want to use Logger directly, as there might be a lot of those, there needs to be another small abstraction.
This is an idea for another pull-request.

src/VM/Handlers/Opcode8044Handler.cpp
phobos2077: Good catch.

src/VM/Handlers/Opcode8046Handler.cpp
janisozaur: Then possibly have a new level in logger, say `Logger::script`. It may need the ability to be toggled on/off regardless of other states.

phobos2077: No, that's not a good idea. Currently we have *levels* and *subsystems* for log events. Different static methods in Logger indicate different event levels. All script warnings are *info* at best, but may use "scripts" subsystem. See this: https://github.com/falltergeist/falltergeist/issues/288.

Issue #296, opened 2015-02-17
Missing cursor after invoking "take from" menu


vvbalashoff:
Easy way to produce:
1. start map arvillag
2. point cursor on a chest in a hut above.
3. try to use it.
Voila! No cursor. And there is NO ESC key working to get rid of it.

phobos2077:
Looks like this is a bug of Container dialog. I'll update it.

PR #305, opened 2015-02-19
State::CursorDropdown refactoring

phobos2077:
Refactoring of CursorDropdown in terms of Mouse cursor changes, state life cycle and icons rendering.

This should fix issues: https://github.com/falltergeist/falltergeist/issues/146 and https://github.com/falltergeist/falltergeist/issues/296.

### DIFF
--- a/src/Game/Game.cpp
+++ b/src/Game/Game.cpp
@@ -174,11 +174,11 @@ void Game::run()
         think();
         render();
         SDL_Delay(1);
-        while (!_statesForDelete.empty())
+        for (auto state : _statesForDelete)
         {
-            delete _statesForDelete.back();
-            _statesForDelete.pop_back();
+            delete state;
         }
+        _statesForDelete.clear();
     }
     Logger::info("GAME") << "Stopping main loop" << std::endl;
 }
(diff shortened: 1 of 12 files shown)


src/Game/Game.cpp
```diff
@@ -174,11 +174,11 @@ void Game::run()
...
         render();
         SDL_Delay(1);
-        while (!_statesForDelete.empty())
+        for (auto state : _statesForDelete)
         {
-            delete _statesForDelete.back();
-            _statesForDelete.pop_back();
+            delete state;
         }
+        _statesForDelete.clear();
```
alexeevdv: Since all objects in vector are already deleted... i think we can skip this "clear" stage :)

phobos2077: We must clear this vector, because it doesn't get cleared anywhere else.

alexeevdv: It will be cleared when vector will be deleted by destructor

alexeevdv:
:+1:

src/Game/Game.cpp
phobos2077: Am I missing something here? _statesForDelete is not a local variable, it's a class member, it will only be cleared when Game is destroyed. But states need to be deleted at the and of every frame. If I don't clear it, it will try to delete object by pointer which was already deleted in previous frame and crash.

alexeevdv: Yep, you are right. I missed that :)

Issue #252, opened 2015-02-09
Jittery walk animation


phobos2077:
1) When dude walks, it looks like not all animation frames are cycling.
2) If you click on tile while dude is already walking, animation should not be "reset"
3) if dude walks in one direction and then you order him to opposite direction - he "teleports" a few hexes

http://screencast.com/t/EFmjDqeWgk

phobos2077:
A correction, point 2 is not valid. Just checked the original, it resets. We could probably improve this by continuing animation when appropriate instead of resetting it, but this is for another issue (improvement over original).

PR #308, opened 2015-02-19
Fix walking and running animations

phobos2077:
1) Fixed incorrect frame offset applying after switching to next hex.
2) Tweaked action frame numbers to make seamless walking and running.
3) Added CritterObject::stopMovements() to immediately stop critter at their current hex.
4) Animations will continue from the same frame after changing direction during movement.

Fixes https://github.com/falltergeist/falltergeist/issues/252


alexeevdv:
:+1:

PR #319, opened 2015-02-28
Egg, part 1

isage:
Base logic are in place and works.
Still needs render re-factoring to be able to work with animated textures.
Still needs flags from map-files to work properly.


phobos2077:
:+1:

PR #321, opened 2015-03-02
New library version && Animation rendering bug fix

alexeevdv:


src/VM/VM.cpp
```diff
@@ -63,34 +63,23 @@ std::string VM::filename()
...
-    try
-    {
-        _script->function(name);
-    }
-    catch (libfalltergeist::Exception &e)
-    {
-        return false;
-    }
-    return true;
+    return _script->procedure(name) != 0;
```
phobos2077: Campare with zero? Maybe objptr will be better?

alexeevdv: Yes. But i still see no difference between zero and zero pointer :)

phobos2077: It will look better. I tend to use zeros when I'm working with some numbers :)

PR #335, opened 2015-03-15
Mapflags

### LINKED ISSUE #318: Object flags in MAP
We should read and set object flags, light values, etc. from MAP and set it for object.
They override values from prototype.

isage:
1. read object flags from map and override flags from proto
2. read light values from map and override values from proto
3. read/set "flat" flag and render flat objects first.

Closes #318

### DIFF
--- a/src/Game/Object.cpp
+++ b/src/Game/Object.cpp
@@ -549,7 +549,7 @@ unsigned int GameObject::lightRadius() const
     return _lightRadius;
 }
 
-void GameObject::parseFlags(unsigned int flags)
+void GameObject::setFlags(unsigned int flags)
 {
     setFlat(flags & 0x00000008);
     setCanWalkThru(flags & 0x00000010);
@@ -566,7 +566,7 @@ void GameObject::parseFlags(unsigned int flags)
 
 }
 
-bool GameObject::flat()
+bool GameObject::flat() const
 {
     return _flat;
 }
--- a/src/Game/Object.cpp
+++ b/src/Game/Object.cpp
@@ -551,6 +551,7 @@ unsigned int GameObject::lightRadius() const
 
 void GameObject::parseFlags(unsigned int flags)
 {
+    setFlat(flags & 0x00000008);
     setCanWalkThru(flags & 0x00000010);
     setCanLightThru(flags & 0x20000000);
     setCanShootThru(flags & 0x80000000);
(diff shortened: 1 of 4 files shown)


src/Game/Object.cpp
```diff
@@ -529,5 +529,52 @@ unsigned short GameObject::lightOrientation() const
...
+{
+    _lightRadius = radius;
+}
+
+unsigned int GameObject::lightRadius() const
+{
+    return _lightRadius;
+}
+
+void GameObject::parseFlags(unsigned int flags)
```
phobos2077: Maybe it's better to name it *setFlags* and change type to uint32_t ?

src/Game/ObjectFactory.cpp
```diff
@@ -308,18 +308,7 @@ GameObject* GameObjectFactory::createObject(unsigned int PID)
     }
     object->setPID(PID);
     object->setFID(proto->FID());
-    object->setCanWalkThru(proto->flags() & 0x00000010);
-    object->setCanLightThru(proto->flags() & 0x20000000);
-    object->setCanShootThru(proto->flags() & 0x80000000);
-
-    object->setTrans(GameObject::TRANS_DEFAULT);
-    if (proto->flags() & 0x00004000) object->setTrans(GameObject::TRANS_RED);
```
alexeevdv: So.... there are no more flags from PRO file? I thought that flags should be overrided by MAP file, not replaced :) Some objects are created directly from scripts using PRO files

src/Game/Object.cpp
alexeevdv: +1 for *setFlags*
We don't need uint32_t here, cause we performing using binary AND operation on flags value

src/Game/ObjectFactory.cpp
phobos2077: He called parseFlags() below, which does the same as the replaced code, as far as I can see.

src/Game/Object.h
```diff
@@ -182,6 +185,17 @@ class GameObject : public EventEmitter
...
+
+    unsigned int lightIntensity() const;
+    virtual void setLightIntensity(unsigned int intensity);
+
+    unsigned int lightRadius() const;
+    virtual void setLightRadius(unsigned int radius);
+
+    virtual void parseFlags(unsigned int flags);
+
+    bool flat();
```
alexeevdv: const :)

src/Game/ObjectFactory.cpp
alexeevdv: Yep. Missed that

src/Game/Object.cpp
```diff
@@ -529,5 +529,52 @@ unsigned short GameObject::lightOrientation() const
...
+{
+    _lightRadius = radius;
+}
+
+unsigned int GameObject::lightRadius() const
+{
+    return _lightRadius;
+}
+
+void GameObject::setFlags(unsigned int flags)
```
phobos2077: how about fixed-width data type for flags?

alexeevdv: There is no need for this as far as i can see

PR #342, opened 2015-03-25
Proper orientationTo inmplementation

isage:


src/PathFinding/Hexagon.cpp
```diff
@@ -146,22 +147,40 @@ void Hexagon::setInRender(bool value)
...
+    unsigned int result = 0;
+
+    if (dx)
     {
-        return dy > 0 ? 0 : 3;
+        // triginometry magick.
+        // basically, we try to find angle to second hex in circle, where first hex is center
+        // and then find out to which of 60� slices it belongs
+
+        double degree = atan2((double)-dy, (double)dx) * 180.0 * 0.3183098862851122; //  180 * 1/PI
```
phobos2077: "nice" constants...

isage: There's non-standart M_1_PI defined in math.h, but it differs in value from original exe.
It's the only place it's used, so there's no need move it somewhere as const double.
I even left the comment describing it, duh.

alexeevdv: http://habrahabr.ru/company/abbyy/blog/252871/

phobos2077: So why not just use 1/M_PI and let compiler worry about constant folding?

isage: M_PI is non-standard extension: it may be defined, may be not.
Same as M_1_PI = 1/M_PI
And once again - it slightly differs in value from original binary. And while i don't know why - i will use value from binary.

phobos2077: That's just seem strange to me. You can use _USE_MATH_DEFINES and #ifndef block, to make sure you have this constant. Alternatively, add it as local constant or something. Surely there should be ways to make it "right" :)

isage:
I fail to see how local constant differs from just a number. Readability?

alexeevdv:
There is no need to use constant. Plain number is good enough. And there is absolutely no need in #ifndef or other macros

PR #344, opened 2015-03-29
TRANS_STEAM flag handling

isage:


isage:
I'll close this for now, but'll leave source branch for reference.
Will be reimplemented in shaders with new render.

PR #346, opened 2015-05-12
Variable length BigCounter counters and add skilldex counters

teep0:
Changed BigCounter counters so that they can be initialized with length because I needed a 3 digit counter for the skilldex counters I added.

### DIFF
--- a/src/State/Location.cpp
+++ b/src/State/Location.cpp
@@ -413,7 +413,7 @@ void Location::think()
     if (_scrollTicks + 10 < SDL_GetTicks())
     {
         _scrollTicks = SDL_GetTicks();
-        unsigned int scrollDelta = 5;
+        unsigned int scrollDelta = 10;
 
         //Game::getInstance()->mouse()->setType(Mouse::ACTION);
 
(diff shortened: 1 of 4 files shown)


src/UI/BigCounter.h
```diff
@@ -40,7 +40,7 @@ class BigCounter : public ActiveUI
     unsigned int _length = 2;
 public:
     enum {COLOR_WHITE = 1, COLOR_RED};
-    BigCounter(int x = 0, int y = 0);
+    BigCounter(int x = 0, int y = 0, unsigned int len = 2);
```
alexeevdv: I think there is no need to use short version of world "length" in code

src/State/Location.cpp
```diff
@@ -413,7 +413,7 @@ void Location::think()
     if (_scrollTicks + 10 < SDL_GetTicks())
     {
         _scrollTicks = SDL_GetTicks();
-        unsigned int scrollDelta = 5;
+        unsigned int scrollDelta = 10;
```
alexeevdv: This must be placed in different commit. Using git branches for every pull request can do the job

PR #355, opened 2015-05-19
add imagebutton disabled textures and enable and disable functions

teep0:
To be used for setting Inventory itemslist, container itemslists, and barter itemslists scroll up and down arrows to disabled arrows textures when scrolling is disabled.

### DIFF
--- a/src/UI/ImageButton.cpp
+++ b/src/UI/ImageButton.cpp
@@ -210,6 +210,8 @@ ImageButton::~ImageButton()
 
 Texture* ImageButton::texture()
 {
+    if(!_enabled) return _textures.at(2);
+    
     if (_checkboxMode && _checked) return _textures.at(1);
 
     if (_hovered && _leftButtonPressed) return _textures.at(1);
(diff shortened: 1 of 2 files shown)


src/UI/ImageButton.cpp
```diff
@@ -219,34 +221,43 @@ Texture* ImageButton::texture()
...
-    if (sender->_checkboxMode)
-    {
-        sender->_checked = !sender->_checked;
-    }
-    if (!sender->_upSnd.empty())
-    {
-        Game::getInstance()->mixer()->playACMSound(sender->_upSnd);
-    }
+    //if(_enabled)
+    //{
```
phobos2077: Why is this commented out?

src/UI/ImageButton.cpp
```diff
@@ -210,6 +210,8 @@ ImageButton::~ImageButton()
 
 Texture* ImageButton::texture()
 {
+    if(!_enabled) return _textures.at(2);
+
```
phobos2077: Maybe we need some constants for those image indexes. Also, don't forget spaces after control flow statements.

PR #358, opened 2015-05-20
Implement TextArea text outlining

teep0:


alexeevdv:
Nice trick. I like it :)

isage:
:+1:

src/UI/TextArea.cpp
```diff
@@ -260,6 +281,39 @@ void TextArea::_calculate()
...
+                outlineSymbolTop->setFont(outlineFont);
+                
+                _symbols.push_back(*outlineSymbolLeft);
+                _symbols.push_back(*outlineSymbolTopLeft);
+                _symbols.push_back(*outlineSymbolBottomLeft);
+                _symbols.push_back(*outlineSymbolRight);
+                _symbols.push_back(*outlineSymbolTopRight);
+                _symbols.push_back(*outlineSymbolBottomRight);
+                _symbols.push_back(*outlineSymbolBottom);
+                _symbols.push_back(*outlineSymbolTop);
```
phobos2077: Maybe reduce code repetition?

isage: Any ideas how? Without making this unreadable (e.g. doing this all in a cycle)
Helper inline function or macros, maybe?

phobos2077: Simply putting repeated stuff in one function would do the job I think, on "client-side" you just tell which offsets you need for each outline. That's DRY principle :)

PR #361, opened 2015-08-12
Fix clang warings

dreamer-dead:
Hi!

Here I'm trying to fix trivial warnings of Clang compiler, mainly related to unsigned types checks.
During this work I realized that duplicated code can be extracted as a method of Texture class.

It seems that my editor removed some trailing whitespaces, and I can revert those changes if you don't like it.

### DIFF
--- a/src/Graphics/Texture.cpp
+++ b/src/Graphics/Texture.cpp
@@ -280,4 +280,27 @@ SDL_BlendMode Texture::blendMode()
     return _blendMode;
 }
 
+bool Texture::blitWithAlpha(Texture* blitMask, int maskOffsetX, int maskOffsetY)
+{
+    if (!blitMask)
+        return false;
+
+    // TODO: Lock both surfaces only once and then use direct pixel access to blend.
+    const auto thisWidth = width();
+    const auto thisHeight = height();
+    const auto maskWidth = blitMask->width();
+    const auto maskHeight = blitMask->height();
+
+    //This is sloooow. But unfortunately sdl doesnt allow to blit over only alpha =/
+    for (unsigned int x = 0, maskX = std::max(0, maskOffsetX); x < thisWidth && maskX < maskWidth; ++x, ++maskX)
+    {
+        for (unsigned int y = 0, maskY = std::max(0, maskOffsetY); y < thisHeight && maskY < maskHeight; ++y, ++maskY)
+        {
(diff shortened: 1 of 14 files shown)


src/VM/OpcodeHandler.h
```diff
@@ -27,6 +27,9 @@
 
 // Third party includes
 
+// Marks handler parameter as unused and prevents compiler warnings.
+#define UNUSED_ARG(arg) (void)(arg)
```
alexeevdv: I don't know why... but i don't like unnecessary defines in our code :) I prefer not to use it where there is no actual need

Oppen: It's also simpler to just remove the argument's name. Programmers understand it's unused and compilers don't complain, so it's probably a convention.

src/Graphics/Texture.cpp
```diff
@@ -280,4 +280,30 @@ SDL_BlendMode Texture::blendMode()
     return _blendMode;
 }
 
+bool Texture::blitWithAlpha(Texture* blitMask, int maskOffsetX, int maskOffsetY)
```
alexeevdv: Nice job. This is good decision to move this code into separate method. We will be able to refactor it later

src/VM/OpcodeHandler.h
alexeevdv: Good point. I think we can just remove unused arguments and leave their names in comments

alexeevdv: We can use something like this:
```C++
/* int movie = */ _vm->dataStack()->popInteger();
```

dreamer-dead: Not exactly, the code 
/* int movie = */ _vm->dataStack()->popInteger();
can lead to warning about unused function value, IIRC

It can be something like
/* int movie = */ (void)_vm->dataStack()->popInteger();

is it better?

alexeevdv: It is definetly better than define :)

dreamer-dead: OK then.

Oppen: Well, those aren't arguments, those are return values.

alexeevdv: You are right. Just variables

dreamer-dead: I mean that they can be considered as some sort of arguments for OpcodeHandlers.
Removed the macro and pushed new version.

PR #363, opened 2015-08-12
Fix possible memory leak (OpcodeHandler object).

dreamer-dead:
If exception will be thrown, then newly created OpcodeHandler object wouldn't be deleted.
std::unique_ptr can easily fix that issue.

### DIFF
--- a/src/VM/VM.cpp
+++ b/src/VM/VM.cpp
@@ -19,6 +19,7 @@
 
 // C++ standard includes
 #include <ctime>
+#include <memory>
 #include <sstream>
 
 // Falltergeist includes
@@ -103,11 +104,10 @@ void VM::run()
         unsigned short opcode;
         *_script >> opcode;
 
-        OpcodeHandler* opcodeHandler = OpcodeFactory::createOpcode(opcode, this);
+        std::unique_ptr<OpcodeHandler> opcodeHandler(OpcodeFactory::createOpcode(opcode, this));
         try
         {
             opcodeHandler->run();
-            delete opcodeHandler;
         }
         catch (VMHaltException& e)
         {


alexeevdv:
Good point. Accepted. Thanks

phobos2077:
Probably we didn't need pointer to an OpcodeHandler in this case (maybe use reference?). Current solution seems overly complex.
http://stackoverflow.com/questions/22146094/why-should-i-use-a-pointer-rather-than-the-object-itself

alexeevdv:
Actually there is no difference. Reference is a constant pointer

Oppen:
Not exactly. You have to be a problematic individual to create a null reference. A null pointer can happen.
It is a matter of abstraction, mostly.

dreamer-dead:
OpcodeFactory creates an object of class inherited from OpcodeHandler (dynamically allocated with new()).
And at this moment we don't know real object type and should destroy it via visual destructor.
How can we use a reference in this situation?

Oppen:
I was just answering the claim they are the same thing. While they are similar, and the same behind scenes, the things the partis promise are different.
I haven't looked at the particular case yet.

dreamer-dead:
Sorry, I was asking phobos2077@ about his idea =)

phobos2077:
@dreamer-dead : I think OpcodeFactory could return OpcodeHandler&, in client code you can try something like:

```
{
   OpcodeHandler& handler = opcodeFactory.create(...);
   // do some stuff
}  // destructor for handler is automatically called

If I'm not mistaken, in this case handler will be created only once in Factory method and then passed just like a pointer, only with nice RAII memory handling (no need to worry about leaks).

```

dreamer-dead:
@phobos2077 , short answer: no, you are getting it wrong.

Our factory should construct and return one of the specific types depending on runtime parameter. Assume that we have types
```
struct Base;
struct A : public Base;
struct B : public Base;
```
In C++ we can return a value from function by value, by reference and by pointer.
Let's consider each of those methods.
- By value
```
Base byValue(bool isA) { 
if (isA) return A {};
return B{};
}
```
It compiles but then it will slice out object and we will have no more than Base.
If Base is an abstract class then code won't even compiles.

- By reference
```
const Base& byRef(bool isA) { 
if (isA) return A {};
return B{};
}
```
Here we will have compiler warning
"warning: returning reference to local temporary object"
It means that program will access some random memory area instead of live object.

- And last method: by pointer
```
Base* byPtr(bool isA) { 
if (isA) return new A {};
return new B{};
}
```
And finally it works as expected, but we should delete the object later.

phobos2077:
@dreamer-dead, OK, you're right. There's also this approach:

```
Factory
{
static std::vector<Base*> ptrs;

Base& create(..) 
{
   ptrs.push_back(/* some factory logic */);
   return *ptrs.back();
}

~Base()
{
   for (p : ptrs) delete p;
}

}

```
We can make factory a singleton (so destructor is called after all handlers are no longer needed), but the ptrs vector will grow constantly, eating up memory. 
Otherwise we could make single pointer variable instead of vector, and issue *delete* every time next object is requested.
At least this way Factory will be responsible for both creating and deleting objects, so you can forget about memory leaks.

I guess other approach is to use proper type of std:: smart pointers, but I never used them and not sure which one :)

dreamer-dead:
I think that OpHandlers shouldn't be allocated by new at all.
Creating and destroying zillions of objects in a second isn't a perfect idea and quite inefficient.
But it seems that you wanted to invent a garbage collection mechanism here.
It's a matter of ownership. In C++ we can use unique_ptr, shared_ptr, intrusive or linked ptrs to declare ownership model. It allows us to clearly understand when and why every object will be destroyed.

Garbage collecting can be helpful, but not in this particular case, I think.
In C++ you always can forget to release resource and to avoid that it's better not deal with raw pointers.
Our factory can return std::unique_ptr<Handler> and then you will be forced to keep it or it will be deleted.

phobos2077:
std::unique_ptr seems good for the task.

PR #364, opened 2015-08-12
Automatic change std::string -> const std::string&.

dreamer-dead:
That's a huge change, but I think it can be useful.
If compiler fails to inline function call, then it will create a temporary copy of a string.
There are STL implementations with std::string optimized for small buffers, and sizeof(std::string) can be greater then 16 for example. Even worse, unwanted string copy can lead to dynamic memory alloc/dealloc.

I'll be glad to discuss this change.


alexeevdv:
Well... i think it is a good idea to pass strings by reference. I'll look into the code later. Thanks

PR #369, opened 2015-08-14
Lua change mouse cursor

dreamer-dead:
Allow to manage mouse cursor from Lua scripts.

### DIFF
--- a/src/Game/Game.h
+++ b/src/Game/Game.h
@@ -65,7 +65,7 @@ class Game
 public:
     ~Game();
     static Game* getInstance();
-    static void export_to_lua_script(Lua::Script* script);
+    static void exportToLuaScript(Lua::Script* script);
 
     void shutdown();
     std::vector<State::State*>* states();
@@ -117,7 +117,7 @@ protected:
     bool _quit = false;
     SDL_Event _event;
     bool _initialized = false;
-    std::unique_ptr<InputController> inputController_;
+    std::unique_ptr<InputController> _inputController;
     void _initialize();
     void _initGVARS();
 private:
--- a/src/Game/Game.h
+++ b/src/Game/Game.h
@@ -32,6 +32,10 @@
 
 namespace Falltergeist
 {
+namespace Lua
+{
+    class Script;
+}
 namespace State
 {
     class Location;
@@ -48,6 +52,7 @@ class UI;
 class TextArea;
 class AnimatedPalette;
 class Settings;
+class InputController;
 
 namespace Game
 {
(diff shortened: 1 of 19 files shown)


src/Input/InputController.h
```diff
@@ -0,0 +1,58 @@
...
+{
+
+namespace Lua
+{
+    class Script;
+}
+
+class Mouse;
+
+class InputController
```
alexeevdv: What is the main idea of InputController class?

dreamer-dead: I think that we shouldn't allow direct access to some engine objects(like mouse) from Lua scripts.
Also we need to do some checks for parameters from scripts.
And this class could do this. We can use more suitable name if you have one.

alexeevdv: For a moment i thought that it is some kind of wrapper for multiple input devices, such as touch pad, mouse or joystick :)

alexeevdv: By the way... i can't see any checks in this class. Actualy we have the same direct access to the Mouse. Am i getting something wrong?

src/Game/Game.h
```diff
@@ -110,10 +117,11 @@ class Game
     bool _quit = false;
     SDL_Event _event;
     bool _initialized = false;
+    std::unique_ptr<InputController> inputController_;
```
alexeevdv: And one more little remark.... we use **_name** instead of **name_** for private and protected members :) I know that codestyle document isn't ready yet, so it is my bad )

dreamer-dead: Strictly speaking identifiers with '_' prefix aren't legal in terms of C++.
All of them are reserved for needs of C++ standard library.
That's why I prefer suffix '_'

alexeevdv: Well... reserved only undescore followed by uppercase letter :)

dreamer-dead: I agree with the answer in topic http://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

"Personally I just don't start identifiers with underscores. New addition to my rule: Don't use double underscores anywhere, which is easy as I rarely use underscore."

Yes, single underscore reserved for global namespace and not for class members.
But it's easier to skip using underscore prefix at all then try to remember all the rules.

alexeevdv: Since there is no problem with undescore in class members i see no reason to change code style it this point

dreamer-dead: It seems reasonable to support the existing code style, and this is minor issue.
So I will fix naming.
But I you will want to introduce a code convention, I recommend you to look at existing one, like Google CC or boost etc, and get the best parts =)

dreamer-dead:
Changed some names.
As an example I filtered all mouse types except three of them and added some debug logging.

PR #370, opened 2015-08-14
Fix const override

dreamer-dead:
Fix for wrong functions override and tons of compiler warnings.
Without const some methods will be hidden but not overloaded.

During this work I moved generic implementation for Counters texture generation.
In future it should be better to use override\const override instead of just virtual.

### DIFF
--- a/src/Graphics/Texture.cpp
+++ b/src/Graphics/Texture.cpp
@@ -306,4 +306,36 @@ bool Texture::blitWithAlpha(Texture* blitMask, int maskOffsetX, int maskOffsetY)
     return true;
 }
 
+// static
+std::unique_ptr<Texture> Texture::generateTextureForNumber(
+    unsigned int number,
+    unsigned int maxLength,
+    Texture* symbolSource,
+    unsigned int charWidth,
+    unsigned int charHeight,
+    unsigned int xOffsetByColor,
+    bool isSigned)
+{
+    // number as text
+    const auto charsCount = maxLength + (isSigned ? 1 : 0);
+    auto texture = std::unique_ptr<Texture>(new Texture(charWidth * charsCount, charHeight));
+
+    // number as text
+    auto number_text = std::to_string(number);
+
+    // Fill counter padding with leading zeroes.
+    if (number_text.size() < maxLength)
+    {
+        number_text.insert(0, maxLength - number_text.size(), '0');
+    }
+
+    for (unsigned int i = 0; i < maxLength; i++)
+    {
+        const unsigned int key = 9 -  ('9' - number_text[i]);
+        const unsigned int x = charWidth * key + xOffsetByColor;
+        symbolSource->copyTo(texture.get(), charWidth * i, 0, x, 0, charWidth, charHeight);
+    }
+    return std::move(texture);
+}
+
 }
--- a/src/Graphics/Texture.cpp
+++ b/src/Graphics/Texture.cpp
@@ -20,6 +20,7 @@
 // C++ standard includes
 
 // Falltergeist includes
+#include "../Base/StlFeatures.h"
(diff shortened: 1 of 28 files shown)


src/Graphics/UI.h
```diff
@@ -75,7 +75,7 @@ class UI
     int _y = 0;
     int _xOffset = 0;
     int _yOffset = 0;
-    Texture* _texture = 0;
+    mutable Texture* _texture = 0;
```
alexeevdv: wow wow wow... what is this for?

alexeevdv:
**override** sprecifier is new for me. I'll try to use it in future. Looks like it is more convinient than **virtual**

src/Graphics/Texture.cpp
```diff
@@ -306,4 +306,36 @@ bool Texture::blitWithAlpha(Texture* blitMask, int maskOffsetX, int maskOffsetY)
...
+    unsigned int maxLength,
+    Texture* symbolSource,
+    unsigned int charWidth,
+    unsigned int charHeight,
+    unsigned int xOffsetByColor,
+    bool isSigned)
+{
+    // number as text
+    const auto charsCount = maxLength + (isSigned ? 1 : 0);
+    auto texture = std::unique_ptr<Texture>(new Texture(charWidth * charsCount, charHeight));
```
alexeevdv: We can define std::make_unique according to http://herbsutter.com/gotw/_102/ and use it in our code. What do you think about it?

dreamer-dead: Unfortunately not now (
It's c++14 featute and my STL library doesn't support it.

alexeevdv: I mean that we can define our own std::make_unique function which can be replaced in future when C++14 is ready

src/Graphics/UI.h
dreamer-dead: I made it mutable to have a const texture() getter.
It is used in width() and height() getters and that's why it should be const.
I understand that this solution isn't perfect (

src/Graphics/Texture.cpp
dreamer-dead: Yes, we can.
It's not a very complex feature.

src/Graphics/UI.h
phobos2077: Why you can't have const texture() getter without "mutable" on the _texture?

alexeevdv: Because in some inherited classes _texture object is created in texture() method. **Const** specifier don't allow this

dreamer-dead: It's a specific to inherited classes but not to UI class.
And there can be a better solution that affects only classes like BigCounter.
I'll spend some more time on that.

dreamer-dead:
I added make_unique helper.
Put it in our namespace and not in std.
We always can add line 'using std::make_unique;' to fix it when c++14 became available.

Oppen:
Why not using namespace std?

dreamer-dead:
Oppen@, 
http://stackoverflow.com/questions/320798/adding-types-to-the-std-namespace

[C++11: 17.6.4.2.1/1]: The behavior of a C++ program is undefined if it adds declarations or definitions to namespace std or to a namespace within namespace std unless otherwise specified.

We will break compilation with c++14 compliant compiler and standard library.
Or we should detect that and do some additional work to avoid collisions.

vvbalashoff:
please rename make_unique to makeUnique

alexeevdv:
@vvbalashoff, make_unique is library function from C++14. There is no need to change it name

dreamer-dead:
Here is my attempt to avoid mutable keyword in base Texture class.
But instead I was forced to add new field for on-demand texture in classes SmallCounter and BigCounter.

Issue #371, opened 2015-08-15
Enum class


alexeevdv:
We need to replace such things (https://github.com/falltergeist/falltergeist/blob/master/src/Game/Object.h#L80) with **enum class**

PR #373, opened 2015-08-17
Issue/371 enum class

phobos2077:
Related to #371 

- Replaced (almost) all enums with enum classes.
- Orientation is a separate class now (wrapper for unsigned char with range validation).
- Moved Subtypes for scenery and items to appropriate classes in inheritance tree.
- Used polymorphism for egg transparency logic instead of complex type & subtype checks in base GameObject class.

### DIFF
--- a/src/Game/WallObject.h
+++ b/src/Game/WallObject.h
@@ -38,6 +38,8 @@ namespace Game
 class GameWallObject : public GameObject
 {
 protected:
+    virtual bool _useEggTransparency() override;
+
 public:
     GameWallObject();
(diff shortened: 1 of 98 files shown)


src/Game/WallObject.h
```diff
@@ -38,6 +38,8 @@ namespace Game
 class GameWallObject : public GameObject
 {
 protected:
+    virtual bool _useEggTransparency() override;
```
dreamer-dead: virtual isn't necessarily if override is used.

PR #376, opened 2015-08-19
Compile with Lua 5.3, and fix memory corruption.

rohit-n:


phobos2077:
Will it still work with Lua 5.1? We need full Luajit compatibility.

alexeevdv:
I don't see anything specific to Lua 5.3 in this commit

phobos2077:
I just wanted to point out that Luajit compatibility was one of the priorities (one of the reasons we choose Lua language to begin with). Need to keep it :)

alexeevdv:
https://github.com/falltergeist/falltergeist/blob/master/CMakeLists.txt#L60

src/Lua/Script.cpp
```diff
@@ -151,8 +151,9 @@ bool Script::get(const std::string& name, bool defaultValue)
     {
         return defaultValue;
     }
-
-    return lua_toboolean(_lua_State, -1);
+    bool ret = lua_toboolean(_lua_State, -1);
```
dreamer-dead: Can we generalize those lines?
Somethings like 

```
#include "LuaBridge/detail/Stack.h"

template <typename T>
T get(lua_State* state) {
    struct StackGuard {
        ~StackGuard() { lua_pop(_state, 1); }

        lua_State* const _state;
    } const guard {state};

    return Stack<T>::get(state);
}

...

return get<bool>(_lua_State);
```
Stack helper from LuaBridge uses luaL_check{number|string} but we already checked type of the stack value and thus we don't need to use lua_to{} API, IMHO

isage:
Why do we even need to manage stack by hand?
>When a C function returns, Lua clears its stack

rohit-n:
@isage I'm new to Lua myself, but not popping the stack was causing an immediate crash on Linux. Valgrind was reporting invalid reads at the last get() in Settings.cpp, and an invalid write when closing _lua_State.

isage:
Can you please provide full stacktrace? Is it only with lua 5.3 or with 5.1 too?

isage:
Oh, sorry, i've missed, that we just reading values from lua script. Yep, we should cleanup stack after we modified it with lua_getglobal.

Issue #307, opened 2015-02-19
Add Point class instead of X, Y


phobos2077:
1. Add class Point(X, Y) with overridden arithmetic operators.
2. Replace setX(), setY(), x(), y() methods with setPosition(), position() in all classes where applicable.
3. Replace duplicated code like this:

```C++
frame->setXOffset(frame->xOffset() - actionFrame->xOffset());
frame->setYOffset(frame->yOffset() - actionFrame->yOffset());
```
With single-line equivalents.

Benefits:
* Cleaner code.
* Less overhead. Let me explain. In some cases, when you call setX() OR setY() on an object, it needs to update some stuff to remain consistent. Now we call setX(), then setY() (two calls) all the time, which may be doing the same thing twice!

PR #378, opened 2015-08-23
Point & Size classes

phobos2077:
Related to #307 
Implemented **Point** and **Size** classes, similar to *QPoint*, *QSize* in Qt.
Replaced *x(), y(), offsetX(), offsetY(), width(), height()* and similar properties with *Point/Size* equivalents in all States and UI elements and some related classes, like LocationCamera, Renderer, Hexagon, etc. It was necessary to introduce those properties globally to take advantage of the shorter code.

No changes were made in Libfalltergeist.
Graphics classes like Texture, AnimationFrame were skipped for now.


src/Input/Mouse.cpp
```diff
@@ -125,39 +136,35 @@ void Mouse::_setType(Cursor state)
             break;
         case Cursor::SCROLL_W:
             _ui = new UI::Image("art/intrface/scrwest.frm");
-            _ui->setYOffset( -_ui->height()*0.5);
+            _ui->setOffset(0, -_ui->size().height() / 2);
```
isage: What if x offset wasn't 0?
Whole point of having separate setX/setY is that we can leave other offset unchanged.

phobos2077: Do you have a better idea? In this case we are setting the absolute position of Image, it is obvious since it right after the constructor.
In case you need to leave other coordinate unchanged, you will have to do:
_ui->setOffset(_ui->offset().x(), .... );

isage: Look how old code worked for transition from S to SW. Are you sure new code works correctly?

src/Point.h
```diff
@@ -0,0 +1,164 @@
...
+    int _width;
+    int _height;
+};
+
+// static class for Rectangle-related function
+class Rect
+{
+public:
+    // checks if given point is located withing the rectangle specified by size
+    static bool inRect(const Point& needle, const Size& size);
```
isage: Why are they static utility function in otherwise empty class? Why not make Rect desendant of Point and Size, and make this member methods (e.g. drop the size, check if "this" intersects with point/rect/etc.) .

src/State/Container.cpp
```diff
@@ -55,8 +55,7 @@ void Container::init()
 
     auto game = Game::getInstance();
 
-    setX((game->renderer()->width()  - 537)/2);
-    setY((game->renderer()->height() - 376)/2);
+    setPosition((Point(game->renderer()->width(), game->renderer()->height()) - Point(537, 376)) / 2);
```
isage: ^this

src/State/Credits.cpp
```diff
@@ -57,8 +57,7 @@ void Credits::init()
 
     Game::getInstance()->mouse()->pushState(Input::Mouse::Cursor::NONE);
     auto renderer = Game::getInstance()->renderer();
-    setX((renderer->width()  - 640)*0.5);
-    setY(renderer->height());
+    setPosition(Point((renderer->size().width() - 640) / 2, renderer->size().height()));
```
isage: ^or this
chose one.

src/State/CritterBarter.cpp
```diff
@@ -50,8 +50,7 @@ void CritterBarter::init()
     setFullscreen(false);
     setModal(true);
 
-    setX((Game::getInstance()->renderer()->width() - 640)*0.5);
-    setY((Game::getInstance()->renderer()->height() - 480)*0.5 + 291);
+    setPosition((Game::getInstance()->renderer()->size() - Point(640, 480)) / 2 + Point(0, 291));
```
isage: ^oh, even third variant.

src/State/Location.cpp
```diff
@@ -254,14 +254,14 @@ void Location::setLocation(const std::string& name)
...
-                auto tile = new UI::Tile(tileNum, x, y);
+                auto tile = new UI::Tile(tileNum, Point(x, y));
                 _floor->tiles()->push_back(tile);
             }
 
             tileNum = mapFile->elevations()->at(_currentElevation)->roofTiles()->at(i);
             if (tileNum > 1)
             {
-                auto tile = new UI::Tile(tileNum, x, y - 104);
+                auto tile = new UI::Tile(tileNum, Point(x, y - 104));
```
isage: I fail to see, how copying even trivial class is better then just passing variables through stack.

src/Input/Mouse.cpp
phobos2077: Which code are you referring to?

src/Point.h
phobos2077: I didn't find any use for a Rectangle class in our code. Only those functions were useful, but no place for Rectangle objects (it would not be convenient if we replace position() and size() pairs with rectangle()...)

src/State/Container.cpp
phobos2077: Missed it, thanks.

src/State/Credits.cpp
phobos2077: This is because we need to divide only X portion of position here... If you have a better idea, please elaborate.

src/State/CritterBarter.cpp
phobos2077: What's wrong with that code? It has different logic than those above.

src/Input/Mouse.cpp
isage: Code in this large switch/case.

src/State/Location.cpp
phobos2077: Is there a big performance issue here? What is your point?
If we use Point classes, we should use them in all UI-related code to take advantage of Point arithmetics.

src/Input/Mouse.cpp
phobos2077: Please be more specific (line #?) I looked through changes in the switch/case and didn't find any difference in logic.

isage: Sorry. I've missed that in each case new ui is created. Nevermind

src/Game/Game.cpp
```diff
@@ -101,7 +101,7 @@ void Game::init(std::unique_ptr<Settings> settings)
     _mousePosition = new UI::TextArea("", renderer()->width() - 55, 14);
     _animatedPalette = new Graphics::AnimatedPalette();
     _gameTime = new Time();
-    _currentTime = new UI::TextArea(renderer()->width() - 150, renderer()->height() - 10);
+    _currentTime = new UI::TextArea("", renderer()->width() - 150, renderer()->height() - 10);
```
dreamer-dead: Can we make a method Size size() const; for Renderer class?

phobos2077: It is there already, only need to use it :)

dreamer-dead: Oh, that method exists.
Can we use it here?

phobos2077: Of course :)

src/Point.h
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * Represents a Point on screen: x and y coordinates, in pixels.
+ */
+class Point
+{
+public:
+    Point() : _x(0), _y(0) {}
+
+    Point(int x, int y) : _x(x), _y(y) {}
```
dreamer-dead: Copy constructor and assignment operator?

src/State/GameMenu.cpp
```diff
@@ -82,37 +83,36 @@ void GameMenu::init()
     // label: save game
     auto saveGameButtonLabel = new UI::TextArea(_t(MSG_OPTIONS, 0), backgroundX+8, backgroundY+26);
     saveGameButtonLabel->setFont(font);
-    saveGameButtonLabel->setWidth(150);
+    saveGameButtonLabel->setSize({150, 0});
```
dreamer-dead: Can we have a constant for this size?
Something like 
```
static const Size kButtonLabel(150, 0);
```

src/Point.h
phobos2077: Why? I thought compiler will create default one for me :)

src/State/GameMenu.cpp
phobos2077: Then we need to refactor A LOT of code, we have those magic number all over the place. Since we plan to re-write all states in Lua, I think we can start using better programming practices there :)

src/LocationCamera.cpp
```diff
@@ -21,93 +21,55 @@
...
-    return _width;
-}
-
-void LocationCamera::setWidth(int value)
-{
-    _width = value;
-}
-
-int LocationCamera::height()
+Size LocationCamera::size() const
```
dreamer-dead: Can we use here signature
```
const Size& LocationCamera::size() const
```
?

phobos2077: For performance? What is your rationale and do you think do the same for other classes?

PR #379, opened 2015-08-23
Make OpcodeFactory to return std::unique_ptr<>

dreamer-dead:
Small improvement.
After discussion in PR #363 we decided to use std::unique_ptr to return an OpcodeHandler from the factory.


src/VM/OpcodeFactory.cpp
```diff
@@ -200,189 +201,190 @@
...
+        case 0x814C: return make_unique<Opcode814CHandler>(vm);
+        case 0x814E: return make_unique<Opcode814EHandler>(vm);
+        case 0x8150: return make_unique<Opcode8150Handler>(vm);
+        case 0x8151: return make_unique<Opcode8151Handler>(vm);
+        case 0x8152: return make_unique<Opcode8152Handler>(vm);
+        case 0x8153: return make_unique<Opcode8153Handler>(vm);
+        case 0x8154: return make_unique<Opcode8154Handler>(vm);
+        case 0x9001: return make_unique<Opcode9001Handler>(vm);
+        case 0xC001: return make_unique<OpcodeC001Handler>(vm);
+        case 0xA001: return make_unique<OpcodeA001Handler>(vm);
```
phobos2077: Is there any way to reduce such repetition of code without macros?

dreamer-dead: We can build a table of factory functions for every opcode.
But these lines will be almost the same when the table will be filled.

phobos2077: I guess I'm too accustomed to dynamic languages (PHP, JS), where I could just put class name in a string and instantiate class from it :)

Issue #324, opened 2015-03-03
Memory corruption in ActiveUI::handle()


phobos2077:
In certain cases current object is deleted after emitEvent(event) call.

Need to refactor event system to prevent such things from happening. Some kind of global (singleton) EventManager may be good choice. EventEmitter objects should still have the same interface for adding event handlers to them (for convenience and code clarity).

PR #381, opened 2015-08-24
Fix event system (proof of concept)

dreamer-dead:
Make event emitting async.
That should fix segfault (#324) while rotating dude on any map.
And also it makes hard to determine was the event handled or not if event1 will lead to emitting event2 and event1.IsHandled flag depends on event2.IsHandled.

### DIFF
--- a/src/Event/Dispatcher.cpp
+++ b/src/Event/Dispatcher.cpp
@@ -23,7 +23,7 @@
 // C++ standard includes
 
 // Falltergeist includes
-#include "../Event/Emitter.h"
+#include "../Event/EventTarget.h"
 
 // Third party includes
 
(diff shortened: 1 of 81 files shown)


src/Game/Game.cpp
```diff
@@ -448,5 +446,19 @@ Audio::Mixer* Game::mixer()
...
+
+}
+
+namespace Event
+{
+// static
+Dispatcher* Dispatcher::getInstance()
+{
+    return &Game::getInstance()->eventDispatcher();
+}
```
phobos2077: Why is this not in Dispatcher.cpp?

src/Event/Dispatcher.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+    if (_scheduledEvents.empty())
+        return;
+
+    decltype(_scheduledEvents) copyOfEvents;
+    swap(copyOfEvents, _scheduledEvents);
+    for (auto& task : copyOfEvents)
+    {
+        task.first->processEvent(std::move(task.second));
+    }
```
phobos2077: Why do we need a copy here?

isage: Because event handler can throw another event and alter queue. We don't want that.

dreamer-dead: To avoid case when _scheduledEvents will be changed during event processing.
If event handler will call emitEvent(event) it will be pushed to _scheduledEvents

src/Game/Game.cpp
dreamer-dead: That's a pseudo-singleton implementation.
I think that knowledge about Game inside Dispatcher.cpp is overkill.

src/Event/Emitter.cpp
```diff
@@ -38,36 +39,40 @@ Emitter::Emitter()
...
-void Emitter::emitEvent(Event* event)
+void Emitter::emitEvent(std::unique_ptr<Event> event)
 {
     if (_eventHandlers.find(event->name()) == _eventHandlers.end()) return;
+
+    Dispatcher::getInstance()->postEventHandler(this, std::move(event));
+}
+
+void Emitter::processEvent(std::unique_ptr<Event> event)
+{
```
phobos2077: Class is called "Emitter", yet it has second responsibility of processing events by name. Are we missing some concept or just need a better name? :)

src/Game/Game.cpp
phobos2077: Why can't you make Dispatcher a singleton of it's own? Otherwise you can just add it as Game property. Singletons are hideous creatures when misused.

src/Event/Emitter.cpp
dreamer-dead: The previous implementation had a capability to group handlers by event name.
I just split it into two phases - delayed post and sync processing.
Do you think that now we should name it like EventSource or something?

src/Game/Game.cpp
dreamer-dead: There are two issues as I see.
First is an access to EventDispatcher instance from Emitter.
As I think it would be better to specify a dispatcher for every Emitter. But it can be a complex task.

Second, is a control over EventDispatcher for Game instance.
If it will be a standalone singleton then it can be hard to manage it's state befor or after Game lifetime.
My solution isn't perfect but it's a compromise.
And I agree with your opinion about singletons.

src/Event/Emitter.cpp
phobos2077: Previously it was a mess :) Need some proper name. 
Look, here is a very similar class https://developer.mozilla.org/en/docs/Web/API/EventTarget

src/Game/Game.cpp
phobos2077: Why do we need EventDispatcher outside of Game lifetime? The game holds all objects and all event-related logic. I think it would be OK to put EventDispatcher inside a Game.

dreamer-dead: It was not so complex to attach a Dispatcher for every Emitter.
Take a look at changes, please.

dreamer-dead:
It seems that this PR still have some problems.
Please don't merge it until I fix the code.

PR #385, opened 2015-08-27
Push zero PID for null objects popped from stack.

rohit-n:
This prevents an error in Opcode8100Handler.cpp when popping a null object pushed earlier in Opcode 8106. This allows the player to talk to more characters in New Reno (tested with Cody, mobsters, Salvatore's men, porn actresses, prostitutes).

### DIFF
--- a/src/VM/Handlers/Opcode8100Handler.cpp
+++ b/src/VM/Handlers/Opcode8100Handler.cpp
@@ -40,8 +40,7 @@ void Opcode8100Handler::_run()
 {
     Logger::debug("SCRIPT") << "[8100] [+] int obj_pid(void* obj)" << std::endl;
     auto object = _vm->dataStack()->popObject();
-    if (!object) _error("obj_pid - obj is NULL");
-    _vm->dataStack()->push(object->PID());
+    _vm->dataStack()->push(object ? object->PID() : 0);
 }
 
 }


phobos2077:
obj_pid with null argument should crash the script (stop current invocation). This is the original behavior of VM.

rohit-n:
Did more research, and that make sense.

Looks like the issue doesn't happen if the player is wearing armor. If they're not, Opcode8106Handler.cpp pushes a null object from critter->armorSlot().

What should be returned if the player has no armor?

phobos2077:
Most probably just a zero. You must understand that original scripts had all kinds of bugs in them :) If you enable debug logs and play for a while, you will see some warnings and errors. 
Maybe the issue is someplace else, or the error handling in original worked somehow differently.

rohit-n:
The issue is definitely caused by Opcode 8106, since it is called immediately before Opcode 8100 when talking to certain characters.

Currently CritterObject::setArmorSlot just changes the player's FID to a jumpsuit if object is NULL.

Maybe when popping the object in 8100, we should first check if it's an ItemObject and set PID to zero if it's NULL?

phobos2077:
When in doubt, look at vanilla engine code, debug scripts in original game. We shouldn't do workarounds. You fix one place, other place will be broken, because you deviated from original behavior.

alexeevdv:
@rohit-n can you provide script name where 8106 is called immediately before 8100?

rohit-n:
An example is talking to Cody in newr1.map. The script is "scripts/nccody.int" and the procedure is talk_p_proc.

alexeevdv:
@rohit-n Thank you. We will look into it

Issue #387, opened 2015-09-06
Optimize libfalltergeist library includes


phobos2077:
Right now if you make even smallest change in any of Lib's headers, 90% of Falltergeist code needs to be recompiled.

Possible solutions:

* Include only needed file types in Falltergeist code, instead of whole Libfalltergeist.h.
* Look through the code and optimize includes where possible (maybe Lib dependency can be moved from .h to .cpp file, etc.)

PR #390, opened 2015-09-06
Optimize lib includes

phobos2077:
Related to #387 

Benefits:
* Reduced dependency from libfalltergeist headers, much less sensitivity to lib's header changes.
* It is more clear what exact file types are used in each source file.


phobos2077:
PR won't compile because of bug in Enums.h of the lib (it didn't include stdint.h which it uses).
A fix for this bug is included in https://github.com/falltergeist/libfalltergeist/pull/38

Issue #383, opened 2015-08-26
Lua integration architecture (discussion)


phobos2077:
I'm not sure if @alexeevdv already made up his mind on how we are going to do this, but I wanted to share my thoughts on the subject.

The way I'd like to see it:
1. Game is initialized in C++ with all basic objects, like an empty state stack, event dispatcher, input, renderer, etc. **No states are defined or created in C++**
2. Then, game executes a single entry point to Lua scripts (something like `"_g.lua"`), a single lua state, global environment for all our logic. That's it, no logic at the engine side!
3. Script has access to constructors for all relevant classes: states, game object, UI controls, events. So it defines all states as a Lua "classes" derived from C++ LuaState class and adds them to the stack via `game.pushState()`.
**Edit:** do we really need LuaState class? We could probably implement State stack entirely on Lua by exposing `render()` method for UIs and game objects... The only question is how to handle events for UI controls...
4. To customize stuff we inherit Lua classes from C++ classes (states, objects, UI's) and override methods like "think", "handle". So each object can define it's logic at each frame.

To implement inheritance we can do something along those lines:
1. As a starting point, take solution from official Lua book: http://www.lua.org/pil/16.2.html
2. The first Lua class will act as a wrapper for userdata (C++) object. Via metatable it will pass calls and property access to underlying object.
3. Second, third, etc. classes will use the object of the previous class in hierarchy as it's __index, so eventually userdata object will be accessed for stuff that wasn't overridden.
4. To pass the underlying object to C++ functions (like `game.pushState()`, or `state:addUI()`) we will use some getter function or property with common name (like `myState.obj`). For example:
`game.pushState(mainMenuState.obj)`
The same will work for other types of object:
`locationState.hexagonGrid.at(2512):attach(superMutantWithMinigun.obj)`
`inventoryState:addUI(game.ui.buttons.RedCircleButton:new(100, 200))`
5. To override C++ functions (like handle() ) with Lua functions, we can add methods like `setHandleFunction()` that will accept Lua function and Lua table as arguments. For example, handle() method in LusState class can look something like that:

```
if (_handleFunction != nullptr) 
{
    _script.pushArgument(_luaTable); // self argument
    _script.callFunction(_handleFunction);
    // get result, etc.
}
```

To organize our code in a clean way, let's use "official" solution: http://www.lua.org/manual/5.1/manual.html#5.3
- We will use strict folder structure.
[...]

PR #401, opened 2015-09-16
Lua Integration

phobos2077:
Related to #383 

- added Inheritable class - to provide ability to "inherit" C++ classes with Lua "classes" (on Lua side, added function that allows to work with Lua class object the same way as C++ object)
- changed how scripted states are invoked, now "main.lua" is responsible for starting the first state
- moved all Lua bindings to separate class (+ additional bindings)
- removed auto-bindings from Script::_initialize() because they are not needed in all cases (configuration, translations, etc.) except actual game scripts
- integrated "middleclass" library for OOP
- used consistent and simple module system for Lua code

As a proof of concept - fully working MainMenu state.


src/Lua/Inheritable.h
```diff
@@ -0,0 +1,104 @@
...
+    luabridge::LuaRef call(const std::string& method, T1 a1, T2 a2, T3 a3) const
+    {
+        return _callInternal(method, [this, a1, a2, a3](luabridge::LuaRef func)
+        {
+            return func(_table, a1, a2, a3);
+        });
+    }
+
+    // calls given method with self and four arguments
+    template <class T1, class T2, class T3, class T4>
```
dreamer-dead: Variadic templates will help!
http://coliru.stacked-crooked.com/a/b53d397337aea4d8

phobos2077: C++ feature? Very cool. Luabridge didn't use it so I thought it's impossible. Looks like it just not made for C++11 :)

PR #405, opened 2015-09-23
State::PlayerPanel replaced with UI::PlayerPanel

alexeevdv:


src/UI/PlayerPanel.cpp
```diff
@@ -0,0 +1,370 @@
...
+            {
+                openPipBoy();
+            }
+            break;
+        case SDLK_z:
+            openPipBoy(); // @TODO: go to clock
+            break;
+        case SDLK_ESCAPE:
+        case SDLK_o:
+            openGameMenu();
```
phobos2077: I think some key presses that are not directly tied to PlayerPanel can be moved to State::Location. Like GameMenu, SaveGame, LoadGame, Exit,

alexeevdv: Yep, but that is the whole new story....

src/UI/PlayerPanel.h
```diff
@@ -0,0 +1,78 @@
...
+    Size size() const override;
+    void render(bool eggTransparency = false) override;
+    void handle(Event::Event *event) override;
+    void think() override;
+
+    void playWindowOpenSfx();
+
+private:
+
+    std::shared_ptr<Image> _background;
```
phobos2077: Why do you need shared_ptr?

src/UI/PlayerPanel.cpp
```diff
@@ -0,0 +1,370 @@
...
+{
+
+PlayerPanel::PlayerPanel() : UI::Base()
+{
+    auto game = Game::getInstance();
+    auto renderer = game->renderer();
+    auto mouse = game->mouse();
+
+    _background = std::make_shared<Image>("art/intrface/iface.frm");
+    _ui.push_back(_background);
```
phobos2077: Seems like it shares ownership with itself :)

PR #411, opened 2015-09-30
Improve loading speed

phobos2077:
- optimized resource manager to cache Dat::Item's previously loaded from DAT file, without trying to find them on disk first (2.5x increase in game loading speed on my setup)

- minor optimizations to UI classes (to cache RGBA textures generated from FRMs via ResourceManager)


alexeevdv:
:+1:

Issue #300, opened 2015-02-18
Word wrap feature for TextArea not implemented


alexeevdv:

phobos2077:
I'll just leave this illustration here :)

![screenshot](http://i.imgur.com/QbDLucX.png)

PR #412, opened 2015-09-30
Word wrap for TextArea

phobos2077:
Related to #300 .

* Implemented word wrap for TextArea.
* Slight optimization (don't regenerate Text Symbols several times on initialization).
* Fixed line interval in Credits.

### DIFF
--- a/src/UI/TextArea.cpp
+++ b/src/UI/TextArea.cpp
@@ -128,6 +128,16 @@ void TextArea::setFont(Font* font)
     _changed = true;
 }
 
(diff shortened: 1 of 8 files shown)


src/UI/TextArea.cpp
```diff
@@ -37,6 +37,8 @@
 // Third party includes
 #include <SDL.h>
 
+using namespace std;
```
alexeevdv: Can you remove this, please?

phobos2077: Done.

alexeevdv:
Good job!

PR #425, opened 2015-10-12
Scrollable message log and dialog reply

phobos2077:
* Added full-featured message log on main interface.
* Added scrolling to reply text on CritterDialog.

TextArea control was heavily changed to accomplish this.


isage:
:+1:

PR #427, opened 2015-10-13
Walking animations rewrite

phobos2077:
A rewrite of walking/running animations. Previous implementation used guessed frame numbers for tile switching. This one uses offset-based approach: 
* After _every_ frame, offset of current animation frame is compared to distance to the next hex in pixels, if it is greater, hex switch is performed and animation offset is subtracted (by value equal to distance between this hex and the next one).
* At turns (when direction is changed before animation cycle ended), current animation frame is "centered" at current hex (this way it is more realistic and matches original).
* At the end of animation cycle, next cycle is offset by value equal to offset of the last frame (this is to achieve seamless walking animation, prevents "limping").

As a bonus, fixed visible weapon in player hands and added take out animation when switching active hands.

### DIFF
--- a/src/Game/CritterObject.cpp
+++ b/src/Game/CritterObject.cpp
@@ -21,6 +21,7 @@
 #include "../Game/CritterObject.h"
 
 // C++ standard includes
+#include <array>
 #include <string>
 
 // Falltergeist includes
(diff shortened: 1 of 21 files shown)


isage:
Are egg working fine after this? It depends on frame offsets.
(It's OK if it's not, btw)
Also :+1:

phobos2077:
Good point, egg is broken (moves by hexes and not by pixels). I'll try to fix it.

phobos2077:
Seems fixed.

alexeevdv:
:+1:

Issue #418, opened 2015-10-05
Timers


phobos2077:
Right now when we need some delayed actions (like when opening CursorDropdown), we create int variables and manually compare them with SDL_GetTicks() on think().

We probably need some abstraction for such timers:
* Create a Timer class with think() method.
* Real-time based timers and ticks-based (think about explosives!) - probably need 2 classes.
* Integration with **add_timer_event** VM opcode (ticks-based timers) - just create timer and attach handler, which will call appropriate VM procedure.
* Use event system for timers:
   - Derive from EventTarget?;
   - Event::Timer class for events;
   - Timer object itself should have handlers for "start", "end".

Thinks to consider:
* Timer is a Game object or independent class?
* Do we attach event handlers directly to Timer or containing object (State, Game::Object?)
* Should Timers be specific to State class or added to any class that needs it (will require to call think() every frame).
* Need to avoid situations when timer handler is called and tries to access object which were deleted.

phobos2077:
A concept of Timers is implemented. Now need to maybe improve it somehow and use everywhere instead of `SDL_GetTicks()` checks.

PR #430, opened 2015-10-13
Timers (concept)

phobos2077:
#418 
Just an example implementation of Timers. There are 2 classes for timers - "Timer" works in real-time (milliseconds) and "GameTimer" in game ticks (which advance only during "real-time" portion of the game). They share the same base class.

Used in 2 places:
* Cursor dropdown, when pressing and holding mouse.
* VM-scripted game timers, added with *add_timer_event* (this can be observed by rats moving about on "klamall" map, for example).

### DIFF
--- a/src/Game/Game.h
+++ b/src/Game/Game.h
@@ -24,14 +24,13 @@
 #include <memory>
 #include <string>
 #include <vector>
-#include <Event/Event.h>
 
 // Falltergeist includes
 #include "../Base/Singleton.h"
 #include "../Game/Time.h"
 
 // Third party includes
-#include "SDL.h"
+#include <SDL.h>
 
 namespace Falltergeist
 {
@@ -41,6 +40,7 @@ namespace Audio
 }
 namespace Event
 {
+    class Event;
     class Dispatcher;
 }
 namespace Graphics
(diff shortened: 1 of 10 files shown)


isage:
Could you up generic timer resolution to nanoseconds?
( for example using https://github.com/falltergeist/falltergeist/blob/master/src/UI/MvePlayer.cpp#L912 )
Or add HiResTimer class.

phobos2077:
I think for general purpose timers there is no point in nanosecond resolution, because in our main loop we have this: 
`SDL_Delay(1);`
Which basically limits FPS to 1000 (if I understand correctly). And since all processing is tied to main loop, nanosecond resolution could only be useful in some vary specific places (can't even think of any, do you?).

isage:
Movies)
Also, SDL_Delay kinda fluctuates.

phobos2077:
I'm not familiar with clock_gettime function. It is not in standard? Also I think if you are dealing with nanoseconds, Timers might be too slow for you because they use Delegate which use std::function - several levels of indirection. Milliseconds-timer is aimed for UI stuff (scrolling, long presses, animations) and basically everything that works inside main loop. Ticks-timer is currently used for "add_timer_event", and other possible use is timed explosives.

PR #438, opened 2015-10-16
Fix minor Clang compiler warnings

dreamer-dead:
Forcing double braces is a kind of bug in Clang, both { and {{ should work due to C++ standard, but single brace generates a warning.
noreturn attribute is a hint for a compiler to not generate a warnings for uninitialized variables.

Please take a look.

### DIFF
--- a/src/VM/OpcodeHandler.h
+++ b/src/VM/OpcodeHandler.h
@@ -27,6 +27,15 @@
 
 // Third party includes
 
+#if defined(_MSC_VER)
+#define NORETURN __declspec(noreturn)
+#elif defined(__clang__) || defined(__GNUC__)
+// Use C++11 attribute, supported by Clang and g++ >= 4.8
+#define NORETURN [[noreturn]]
+#else
+#define NORETURN
+#endif
+
 namespace Falltergeist
 {
 class VM;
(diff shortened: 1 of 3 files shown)


phobos2077:
Why do we need to adjust our code to work around CLang bugs? Double braces doesn't look right to me :)

src/VM/OpcodeHandler.h
```diff
@@ -27,6 +27,15 @@
...
 // Third party includes
 
+#if defined(_MSC_VER)
+#define NORETURN __declspec(noreturn)
+#elif defined(__clang__) || defined(__GNUC__)
+// Use C++11 attribute, supported by Clang and g++ >= 4.8
+#define NORETURN [[noreturn]]
+#else
+#define NORETURN
+#endif
```
phobos2077: Since we fully use C++11, why don't just always use `[[noreturn]]` ?

dreamer-dead: MSVS doesn't support it yet, IIRC.

dreamer-dead:
Double braces is another standard way to do it.
If it will fix some warnings, why not use it?

src/VM/OpcodeHandler.h
phobos2077: Why care about MSVS if it doesn't support C++11? It probably won't compile anyway, with or without this macros.

phobos2077:
I suggest replacing NORETURN macro with C++11 standard attribute.

alexeevdv:
No way...We will not fix WARNINGS using MACRO :) Sorry

phobos2077:
Well [[noreturn]] is pretty valid option in C++11, it's not a macro. Double braces make sense too, if you study the standard...

alexeevdv:
Yep, but it has nothing to do with this particular pull request

PR #446, opened 2015-10-27
Don't stop music after destroying MainMenu.

rohit-n:
This allows music to play when starting a new game.


isage:
But why? Original game stops music when changing menus. Why shouldn't we?

rohit-n:
The problem is that MainMenu is destroyed after you load the map, immediately stopping music. Maybe there's a better way?

isage:
That's a bug. MainMenu should be destroyed prior to starting level.

rohit-n:
How about now? Just delete all states set for deletion after setting a new state.

phobos2077:
This doesn't fix similar issues when you first pop state and then push another state. Also, `_statesForDelete` was designed so states would only be destroyed at the end of the frame, to avoid possible segfaults. 
What you probably want is to use appropriate event handlers for states. State constructors and destructors should ONLY allocate/deallocate resources for the state. They should not alter global state (like music being played);

rohit-n:
If destructors shouldn't alter global state, should I remove the stopMusic call in MainMenu's destructor? That was originally my plan.

phobos2077:
Try moving it to `onStateDeactivate` instead. Original behavior should be retained.

rohit-n:
Didn't work. onStateDeactivate stops music when going to character selection menu, and onStatePop (when "pop" is emitted to popHandler) is called after the location music plays.

rohit-n:
How about now? Stop music when MainMenu pops, play music when Location activates.

PR #447, opened 2015-11-01
Reset dialog line offset when selecting an answer.

rohit-n:
This prevents a bad alloc segfault when transitioning
from a multi-scroll dialog question to single scroll
while the line offset > 0.

To test, talk to Slim Picket in klamall.map, and press "1" two times. Scroll to the bottom of the response text, then press "1" again.

### DIFF
--- a/src/State/CritterDialog.cpp
+++ b/src/State/CritterDialog.cpp
@@ -134,6 +134,7 @@ void CritterDialog::setQuestion(const std::string& value)
 {
     auto question = getTextArea("question");
     question->setText(std::string("  ") + value);
+    question->setLineOffset(0);
 }
 
 void CritterDialog::onAnswerIn(Event::Mouse* event)


src/State/CritterDialog.cpp
```diff
@@ -134,6 +134,8 @@ void CritterDialog::setQuestion(const std::string& value)
 {
     auto question = getTextArea("question");
     question->setText(std::string("  ") + value);
+    question->setLineOffset(0);
+    Game::getInstance()->mouse()->setState(Input::Mouse::Cursor::BIG_ARROW);
```
phobos2077: Why mouse state change?

rohit-n: Purely aesthetics. If you are currently viewing a multi-scroll question, the mouse cursor will show as an up or down arrow to indicate this. If you then press a number on the keyboard and a single-scroll question appears, the cursor will not change, so I explicitly set it.

I just checked vanilla FO2 and found that this isn't original behavior, so I removed that line.

PR #448, opened 2015-12-13
Work towards support for Visual Studio 2015.

8BitPimp:
Hi there!
I was working on my own fallout reverse engineering project, when I saw the amazing work you folk are doing here.

This pull request works towards support for Visual Studio 2015:
Included in this patch:
- Have split CrossPlatform into separate implementations for Unix hosts and Windows hosts; PlatformUnix and PlatformWindows.
- PlatformWindows uses FindFirstFile() from the windows api, in place of dirent.h which doesnt exists for MSVC.
- I had to rework the use of make_unique<> since MSVC implements this.  Faltergeists implementation is now not used when building on MSVC.
- I have refactored clock_gettime() into platform as CrossPlatform::getTime().  This is implement for both unix (using clock_gettime) and windows (using QueryPerformanceCounter).
- I had to explicitly include many of the headers that were being implicitly included on linux.

Please note: the above does not fix MSVC builds entirely and work is still required.  I thought however that I would stop before this pull request got too large.

Thanks


phobos2077:
Thank you, we will review this later. 
Btw, about reverse engineering, have you heard about *sfall*?

src/Platform/PlatformUnix.cpp
```diff
@@ -16,6 +16,7 @@
  * You should have received a copy of the GNU General Public License
  * along with Falltergeist.  If not, see <http://www.gnu.org/licenses/>.
  */
+#if defined(__unix__) || defined(__APPLE__) || defined(__linux__)
```
janisozaur: In the other project I contribute to, I was told for OS X you need `(defined (__APPLE__) && defined(__MACH__))`

I also believe `PlatformPosix` would be a better name for the file.

src/Platform/PlatformWindows.cpp
```diff
@@ -0,0 +1,293 @@
...
+ *
+ * Falltergeist is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Falltergeist.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#if defined(_WIN32) || defined(_MSC_VER)
```
janisozaur: Is there anything MSVC-specific? this check would prevent people from building with mingw.

src/Platform/PlatformUnix.cpp
8BitPimp: Hi, Janisozaur
Thanks for the info, and I agree with you on the 2nd point also.
I will revise the pull request to fix both these points.

8BitPimp: HI Phobos,
I have heard of sfall yes.  Its a really nice project, however it seems to focus on patching and extending, rather then reimplementing, which I was aiming to do.
Using IDA I was converting fallout back into C code, 1 to 1, simmilar to what KeeperFX did to DungeonKeeper.  Progress was really slow however.
It really seems more usefull to collaborate with you folk on this project!

janisozaur: @8BitPimp just a side note about what you said: take a look at https://github.com/OpenRCT2/OpenRCT2 then.

Main dev's screencast: https://www.youtube.com/watch?v=0AMIE2sj2AA

src/Platform/PlatformWindows.cpp
8BitPimp: MinGW should define _WIN32, so this code should be ok.  I definately dont want to break MinGW support.

src/Platform/PlatformUnix.cpp
8BitPimp: Wow!  Thats amazing, thanks, I hadnt seen this before :)

Oppen: Does this look necessary? Shouldn't the build system handle the conditional build of the whole file there?

8BitPimp: I tend to agree with you Oppen, it seems like something CMake should be doing.  From what I could see however, CMake is currently globbing all files recursively, so it will grab both files.  Im not too strong with CMake, so if thats a better aproach perhaps someone could assist?

alexeevdv: It has nothing to do with CMake. C++ preprocessor will process this file

Oppen: The recommended approach (by the CMake folks, I mean) is to avoid globbing, but one less troublesome option could be to avoid to use it recursively. You glob src, you glob the folders where everything should be built unconditionally, but you don't glob over Platform, where conditions must be met for the files to be built.

Oppen: @alexeevdv, why? The question is precisely if it should. We are guarding inside the file for ignoring it whole. That's CMake's job. If it's a piece of code inside that is platform specific, it makes sense. If it's the whole file platform specific, I don't believe so.

alexeevdv: @Oppen, because CMake is not a compiler. It doesn't process defines in C++ files

Oppen: @alexeevdv, no CMake is a build system. For organizing which files to build. If the preprocessor will parse the file just to decide it will not be built because you are not building for the platform in which that file is relevant, you don't need it to even get to the preprocessor, you already know from the outside you don't want to build that file.
The file is clearly names "PlatformUnix", and that's because all of the file is Unix specific. You don't need to get to the preprocessor stage to know you don't want it built on Windows and you want it built on POSIX based OSes. CMake already knows that.

alexeevdv: Good point. This  can be done using list(REMOVE_ITEM) as described here: http://stackoverflow.com/a/16450083/3137053

8BitPimp:
I have updated my pull request with some of the changes mentioned:
- PlatformUnix has been renamed to PlatformPosix. (janisozaur)
- CMake is now responsible for selecting either PlatformWindows or PlatformPosix. (alexeevdv)

src/Base/StlFeatures.h
```diff
@@ -72,6 +73,10 @@ make_unique(std::size_t n)
 template<class T, class... Args>
 inline typename Detail::UniqueIf<T>::KnownBound
 make_unique(Args&&...) = delete;
+#else
+// Visual Studio 2015 implements make_unique
```
phobos2077: What about older VS versions? This condition assumes that any VS supports make_unique.

alexeevdv: @phobos2077, Correct me if i wrong.... but older VS versions don't support C++11

8BitPimp: The only two versions of VS that support C++11 are 2013 and 2015.  According to MSDN, they both implement make_unique.  I have only actualy tried compiling with Visual Studio 2015 however.

dreamer-dead: IIIRC, MSVS2010 was able to support of some C++11 features (but not all, of course).
Also there was MSVS2012.
But make_unique is a C++14 feature, and it supported in MSVS2013 and MSVS2015 only.
It will be better to check the value of _MSC_VER macro for compiler version.

src/Font.cpp
```diff
@@ -28,7 +28,7 @@
 
 namespace Falltergeist
 {
-using Base::make_unique;
+using namespace Base;
```
dreamer-dead: Why do you need all of the Base namespace content?

src/Platform/PlatformPosix.cpp
```diff
@@ -378,4 +335,29 @@ std::vector<std::string> CrossPlatform::getDataPaths()
     return _dataPaths;
 }
 
+void CrossPlatform::getTime(TimeInfo * out) {
+    if (!out)
+        throw Exception("nullptr passed into getTime");
+#if defined(__MACH__)
+    static mach_timebase_info_data_t timebase;
+    mach_timebase_info(&timebase);
```
dreamer-dead: Static variable will be filled every time.
Do you expect that? If yes, then why do you use `static` ?

src/Platform/PlatformWindows.cpp
```diff
@@ -0,0 +1,290 @@
...
+{
+    if (_falloutDataPath.length() > 0)
+        return _falloutDataPath;
+    Logger::info() << "Looking for Fallout data files" << std::endl;
+    std::vector<std::string> directories;
+    directories.push_back(getExecutableDirectory());
+
+    for (auto &directory : getDataPaths())
+    {
+        directories.push_back(directory);
```
dreamer-dead: Can we just use `std::vector::insert` ?

src/Platform/PlatformWindows.cpp
```diff
@@ -0,0 +1,290 @@
...
+{
+    char path[MAX_PATH];
+    SHGetFolderPath(nullptr, CSIDL_APPDATA | CSIDL_FLAG_CREATE, NULL, 0, path);
+    return std::string(path) + "/falltergeist";
+}
+
+std::vector<std::string> CrossPlatform::getDataPaths()
+{
+    std::vector<std::string> _dataPaths;
+    _dataPaths.push_back(getConfigPath());
```
dreamer-dead: Can we use std::initialiser_list to avoid unecessary allocations in std::vector?
Smth. like `return { getConfigPath() };` ?

src/UI/MvePlayer.cpp
```diff
@@ -908,9 +890,9 @@ void MvePlayer::think()
 {
     if (!_timerStarted) return;
 
-    struct timespec ts;
-    clock_gettime(CLOCK_MONOTONIC, &ts);
-    uint32_t nsec = (ts.tv_nsec - _lastts.tv_nsec);
+    struct TimeInfo ts;
```
dreamer-dead: You don't need to write `struct SomeName var;` in C++.
Also you should consider returning of this structure by value.
Sometimes it can be more efficient than passing a pointer and writing to it.

src/UI/MvePlayer.cpp
```diff
@@ -908,9 +890,9 @@ void MvePlayer::think()
...
+    CrossPlatform::getTime(&ts);
+    uint32_t nsec = (ts._nano - _lastts._nano);
```
dreamer-dead: Type of the `TimeInfo::_nano` is `unsigned long long` which is std:: uint64_t I suppose.
And `nsec` variable should be the same.
Use `auto` keyword to keep it simple.

src/Platform/PlatformWindows.cpp
```diff
@@ -0,0 +1,290 @@
...
+    return _dataPaths;
+}
+
+void CrossPlatform::getTime(TimeInfo * out) {
+    if (!out)
+        throw Exception("nullptr passed into getTime");
+
+    // Get ticks per second
+    LARGE_INTEGER Frequency;
+    if (QueryPerformanceFrequency(&Frequency) == FALSE)
```
dreamer-dead: It will be better to call this function only once.

See https://msdn.microsoft.com/ru-ru/library/windows/desktop/ms644905(v=vs.85).aspx
`Retrieves the frequency of the performance counter. The frequency of the performance counter is fixed at system boot and is consistent across all processors. Therefore, the frequency need only be queried upon application initialization, and the result can be cached.`

src/Platform/PlatformPosix.cpp
```diff
@@ -378,4 +335,29 @@ std::vector<std::string> CrossPlatform::getDataPaths()
...
+    uint64_t time;
```
dreamer-dead: Why not a one-liner?
`uint64_t time = mach_absolute_time();`

src/Platform/Platform.h
```diff
@@ -32,6 +32,11 @@
 namespace Falltergeist
 {
 
+struct TimeInfo {
+    unsigned long long _sec;
```
dreamer-dead: It will be better to use something like std::uint64_t here, IMHO.

isage:
Sorry this wasn't merged. I'm currently plan on working towards VS compatibility.
This pull-request, however, still won't be merged, as i intend to do some things differently.
Thank you for your work, it wasn't for nothing, as i'll cherry-pick and use some of it.

Also, i propose we drop our own make_unique implementation, as all modern compilers support it.
And we don't need to support old versions of VS, because it's free.

alexeevdv:
+1 for std::make_unique

isage:
This can be closed now, i think.

PR #450, opened 2016-02-27
Glorious modern-GL render

isage:
So, here it is.
* New modern opengl render with shaders.
* Working roofs, with roof transparency checking and blocking events.
* Critter outlines (for battle) and cursor outline for red hex.
* Red hex cursor now rendered after tiles, with outline rendered after all objects. Same as in original.
* From now on, all images or frames in atlases (except tilemaps) have 1px transparent border around, this is required for correct outlines.
* Animated palette and object transparency (energy, steam, etc.) are moved into shaders.
* Egg transparency is moved to shader. This is *much* faster, than previous software per-pixel solution.
* Fonts with outlines, outlines in shaders, instead of drawing font several times.
* Support for .FON fonts.
* Global illumination. Maps/objects are lighted according to map light script value (opcode 80E9 implemented).
* Render cycle optimizations: we now store and iterate over array of objects, instead of array of all hexes. This is faster.
* General optimizations: no more texture copying, resampling, etc. Resizing is done by render.

TODO:
Of course, we're all people, so this maybe not ideal from OOP standpoint, so refactoring are welcome.
Also, i've not always followed coding standards about header includes. I'm a little lazy to fix it now.
Sadly, for now we require OpenGL 3.3 or better. I intend to add support for lower versions later, as this requires writing separate render-path and shaders, and i want for new render to be merged first.
And, ofcourse, there's still lots of things to do: like framebuffers, for postprocessing effects, or like object lights.


janisozaur:
wow

Oppen:
First of all, I'll quote janisozaur and say 'wow'.
Secondly, a few questions and suggestions:
Question number one is if this is supposed to be optional or if you completely replaced the original implementation (I'd check it out by myself, but I'm not too versed in the inners of the project and there are too many commits, easier to just ask). I believe it would be a good idea, and you'd need to write the infrastructure anyway to support older versions of OpenGL, to allow for runtime setting of the renderer, with a fallback to software.
Suggestion number one, I've seen you used floats (a commit mentions it), it might be convenient to alias the type to allow for fixed point, since that would allow you to later create an OpenGL ES version, which AFAIK (at least up until version 2) is fixed point.
Considering the target games are relatively easy on resources, supporting low end devices through GL ES and similar might be desirable.

I might later check if I can suggest anything, but I'm not an OpenGL expert nor an acknowledgeable member of the project, so I doubt I will be able to give you any break-through insight.

isage:
That's completely replaces SDL render, as it's just impossible to do some things with SDL.
There's no software fallback (and won't be).
"Floats" in the commit have nothing to do with render, there was a bug in sliders code.
We do, however, use floats in shaders, but GLES render would require different set of shaders anyway.
As for GLES, it's fixed point only in Common Lite profile. And i'm not really sure if we really need (and can) to support GLES<2.0. 
Btw, target games are easy on resources, yes. But our engine sadly not (yet).

Oppen:
Fair enough.

phobos2077:
WOW indeed. What about platform support after these changes? Will we face any issues when porting to Android/iOS/etc?

isage:
Not really, same issues as for OpenGL 2.1 support: different shaders and slightly different renderpath

PR #454, opened 2016-03-14
Bump version and update readme

isage:
And also quick-fixed two annoying bugs.


isage:
Hm. Don't merge it yet. I have couple of optimizations coming, which will drastically improve performance, i hope. Then we can call it new shiny version :)

isage:
Okay, now i'm pretty happy with results, so, good to merge.

alexeevdv:
Very good :) Thank you

PR #459, opened 2016-03-19
Implement dynamic enable/disable inventory scroll buttons

g4th:
Enabling/disabling of inventory scrolling partially implemented, the only thing that needs to be added is updating when an item is removed or added to the list. Also add feature to disable and enable ImageButtons.

Edit:
scroll buttons update when item is removed and added now.


src/State/Inventory.cpp
```diff
@@ -308,6 +330,47 @@ void Inventory::onScrollDownButtonClick(Event::Mouse* event)
...
+            toggleScrollDownBtn(false);
+        }
+        if(inventory->canScrollUp())
+        {
+            toggleScrollUpBtn(true);
+        }
+    }
+}
+
+void Inventory::toggleScrollUpBtn(bool toggle)
```
isage: 1) if that's "toggle", it should do exactly that, toggle, no parameters needed. Otherwise it should be named something like setEnabled or setState
2) please, don't abbreviate, Btn should be Button

src/State/Inventory.cpp
```diff
@@ -308,6 +330,47 @@ void Inventory::onScrollDownButtonClick(Event::Mouse* event)
...
+{
+    auto scrollUpBtn = dynamic_cast<UI::ImageButton*>(getUI("button_up"));
+    auto scrollUpBtnDisabled = dynamic_cast<UI::ImageButton*>(getUI("button_up_disabled"));
+    if(toggle)
```
isage: This can be just
    scrollUpBtnDisabled->setEnabled(!toggle);
    scrollUpBtn->setEnabled(toggle);

isage:
Otherwise looks good.

isage:
Hm, also, why not make disabled state just an Image instead of ImageButton?

g4th:
I was uncertain if the disabled scroll buttons had a sound effect on click in the original, so I just made them ImageButtons in case later on I find out they do.

isage:
They don't.

g4th:
Would it be better if it were an image? Would have to add enable and disabling of Images also.

isage:
Yeah, i think they better of as Image.
Btw, there's "visible" flag and getters/setters in Ui/Base, you should use that.

g4th:
Disabled scroll buttons now are images. Also stopped handling mouse events of non-visible images.

isage:
First commit was actually unneeded, as State in render loop ignores UIs with visible=false.
Looks good to merge. Let's see what @alexeevdv has to say.

g4th:
> First commit was actually unneeded

Fixed

alexeevdv:
Looks good. Thank you

Issue #155, opened 2014-10-09
correct font in Options (SettingsMenuState)


vvbalashoff:
Now font of labels "combat difficulty" and some others is incorrect

phobos2077:
Also need to correct TextArea positions.

PR #463, opened 2016-03-25
Correct SettingsMenu label font and SettingsMenu label placement

g4th:
overlaid_settings
falltergeist_fixed_settings_menu

Correct the Settings header font and placement of labels in SettingsMenu
#155
Top picture is Settings menu before overlaid with original game's settings menu and bottom picture is settings menu after overlaid with original game's settings menu.

It appears that the original engine inserted extra spaces in the word wrapped texts 'Maximum Blood' and 'Targeting Only'.


g4th:
Inserting an extra space in word wrapped labels makes SettingsMenu pretty much perfect with original settings menu
falltergeist_fixed_settings_menu2

isage:
Nice

PR #470, opened 2016-03-27
Implement spatial scripts

isage:
I've implemented spatial scripts and fixed/implemented/mocked some of VM opcodes.
That's still not enough for compete implementation (well, it is, from spatials side), but it works.

Tested on green goo in arvillag and on spear traps in arcaves.
Spear traps, however, require more work, as they set/advance dude animation in while cycle, so we should call to render inside op_anim (which is unimplemented).
But this is out of scope of spatials implementation.


isage:
While i'm at it, i've also fixed/implemented another couple of opcodes

alexeevdv:
Good job. Now we have traps! )

PR #473, opened 2016-04-04
Fix typo

janisozaur:


isage:
Ugh. That explains why i got strange results sometimes. Nice find.

PR #475, opened 2016-04-30
VisualStudio™ build

isage:
Welp, here it is. I've also fixed couple of undefined-behavior bugs and one bug in 2.1 render.
Due to drop of our own make_unique implementation build now requires gcc 4.9/clang 3.5 or later.
Should be available in any recent distro, though.

I'll post build tutorial(s) after merge.


src/Format/Ini/File.cpp
```diff
@@ -52,7 +52,7 @@ Section& File::section(const std::string &name)
     if (it == _sectionIdxMap.end())
     {
         auto idx = _sections.size();
-        _sections.push_back(std::move(Section(name)));
+        _sections.push_back(Section(name));
```
dreamer-dead: Maybe just 
_sections.emplace_back(name);
?

src/Format/Ini/Value.cpp
```diff
@@ -76,7 +76,7 @@ int Value::toInt() const
     {
         return std::stoi(_value);
     }
-    catch (std::invalid_argument)
```
dreamer-dead: I'm interested, why exceptions often handled by non-const reference?
Will it be changed? =)

src/Format/Ini/File.cpp
isage: Sure

src/Format/Ini/Value.cpp
isage: Not that it's really important, but yep, sure.

src/PathFinding/HexagonGrid.cpp
```diff
@@ -247,7 +245,7 @@ Hexagon* HexagonGrid::hexInDirection(Hexagon* from, unsigned short rotation, uns
 std::vector<Hexagon*> HexagonGrid::ring(Hexagon* from, unsigned int radius)
 {
     std::vector<Hexagon*> result;
-    Hexagon* current = nullptr;
+    Hexagon* current;
```
dreamer-dead: Why this change?
We can just move down this local variable to line 255.

isage: Two consecutive assignments. But yeah, it should be moved down. Done.

isage:
Btw, we also tested this branch on openbsd/freebsd, works there too.

src/ResourceManager.cpp
```diff
@@ -403,8 +402,7 @@ Graphics::Shader* ResourceManager::shader(const string& filename)
         return _shaders.at(filename).get();
     }
 
-    Graphics::Shader* shader = nullptr;
-    shader = new Graphics::Shader(filename);
+    Graphics::Shader* shader = new Graphics::Shader(filename);
 
     _shaders.insert(make_pair(filename, unique_ptr<Graphics::Shader>(shader)));
```
dreamer-dead: It seems that we can use emplace() method like
_shaders.emplace(filename, make_unique<Graphics::Shader>(filename)));

src/UI/MvePlayer.cpp
```diff
@@ -837,8 +819,10 @@ void MvePlayer::_processChunk()
         return;
     }
 
-    for (auto opcode : *_chunk->opcodes())
+    auto opcodes = _chunk->opcodes();
+    for (unsigned int i=0; i < opcodes->size();i++)
```
dreamer-dead: Can we use range-based for loop here? Like 

`for (const auto& opcode : opcodes) ...`

src/Format/Ini/Value.cpp
dreamer-dead: That is a common practice, I see a lot of code with that approach from different developers.
But this is quite strange for me =)

src/ResourceManager.cpp
isage: Yeah, but we still need to return shader.

dreamer-dead: This makes sense, thanks for clarification.

src/UI/MvePlayer.cpp
isage: _chunk->opcodes() returns pointer, so
````
auto opcodes = *_chunk->opcodes();
for (const auto& opcode : opcodes)
````
Strangely, first time i tried, VS croaked with error.

dreamer-dead: Be careful, because

`auto opcodes = *_chunk->opcodes();`
will make a copy of the vector.
`const auto& opcodes = *_chunk->opcodes();`
is a preferred way.

isage: Oh, right. Fixed.

Issue #444, opened 2015-10-20
CritterAnimationId class


phobos2077:
We probably need another layer of abstraction for critter animation IDs. We need to map FID numbers to FRM names, and build FRM names from components (base Art ID, weapon ID, animation ID, etc.).

Suggest to create some helper class with properties `artId`, `weaponId`, `animId`, constructors from full FID (32bit integer) and FRM name (string), and functions to generate those. It will be used by Action classes to build correct FRM names for Animations using components. Right now we directly use portions of strings (eg. for weapon) and it's really messy (see CritterObject class).

Right now we have function that generates FRM name from FID in ResourceManager class, will need to move it.

PR #478, opened 2016-05-20
Simplify and speed up of the FIDtoFrmName function.

dreamer-dead:
This is a first part of the #444 task.
Here I tried to make FIDtoFrmName more clean and fast.
If this will be good enough to merge, I can continue working on that task.

### DIFF
--- a/src/ResourceManager.cpp
+++ b/src/ResourceManager.cpp
@@ -18,12 +18,12 @@
  */
 
 // C++ standard includes
-#include <array>
 #include <fstream>
 #include <sstream>
 #include <iomanip>
 #include <locale>
 #include <memory>
+#include <utility>
 
 // Falltergeist includes
 #include "CrossPlatform.h"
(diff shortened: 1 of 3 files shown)


src/ResourceManager.cpp
```diff
@@ -94,20 +161,27 @@ ResourceManager* ResourceManager::getInstance()
...
-    std::transform(filename.begin(), filename.end(), filename.begin(), ::tolower);
-
     // Return item from cache
     auto itemIt = _datItemMap.find(filename);
     if (itemIt != _datItemMap.end())
     {
         return itemIt->second;
     }
 
+    std::string filenameLowered(filename);
```
isage: Um. Why do this twice? It doesn't make sense to lookup non-lowercased filenames in _datItemMap, because we always use lowercased filenames.

dreamer-dead: The old code does two things:
- Always makes a copy of an input string
- Transforms the string to lowercase

My point is that we can avoid these actions if input string already is in lowercase.
And then do the transformation only if we failed to find the input string.

If you think that we should use lowercased filenames only, then we can delete this step completely.

alexeevdv: Hi! We can remove this step completely for sure :)

PR #482, opened 2016-06-17
Update Readme and add Linux Desktop support

shsorbom:
The readme was missing the zlib dependency and information on global installs.
Installation of a .desktop file is necessary in Linux for Falltergiest to show up in desktop menus.
I did not include the bundled icon because it is the wrong size and filename. I did not wish to tamper with it in case there was a reason for this.


isage:
1) we are not currently accepting packaging pull-requests. Because we are in early alpha WIP stage and it doesn't make sense. However, as it's relatively small and trivial, i'm okay with making exception this time.
2) "Remake of Fallout 2" - it's not a remake. It's opensource engine for it.
3) >zlib1g - a think we should remove distro-specific package names from readme, or create distro-specific sections.

isage:
Also, ignore Travis-CI failure. Looks like we need to find another source of fresh clang packages, meh.

isage:
Also, there's no reason for icon filename/size/etc. It's just a placeholder for now.

shsorbom:
How does adding new commits for a pull request work? do I nee to file another one to fix the comment issue?

isage:
Just add commit to the same branch

shsorbom:
Done.

Issue #486, opened 2016-08-17
Cannot load map from data directory


JanSimek:
Game crashes when loading map from data directory instead of data file.

```
[DEBUG] [RESOURCE MANAGER] Loading file: art/splash/splash6.rix [FROM /home/jsimek/Development/falltergeist/build//master.dat]
[DEBUG] [RESOURCE MANAGER] Loading file: art/intrface/wait.frm [FROM /home/jsimek/Development/falltergeist/build//master.dat]
[DEBUG] [RESOURCE MANAGER] Loading file: color.pal [FROM /home/jsimek/Development/falltergeist/build//master.dat]
[INFO] [GAME] Starting main loop
[DEBUG] [RESOURCE MANAGER] Loading file: premade/combat.gcd [FROM /home/jsimek/Development/falltergeist/build//master.dat]
[DEBUG] [RESOURCE MANAGER] Loading file: art/intrface/actarrow.frm [FROM /home/jsimek/Development/falltergeist/build//master.dat]
[DEBUG] [RESOURCE MANAGER] Loading file: maps/test.map [FROM FALLOUT DATA DIR]

Thread 1 "falltergeist" received signal SIGSEGV, Segmentation fault.
0x00007ffff6a91924 in std::istream::seekg (this=0x7fffffffd280, __off=0, __dir=std::_S_end) at /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/istream.tcc:915
915     /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/istream.tcc: Directory or file does not exist.
(gdb) bt
#0  0x00007ffff6a91924 in std::istream::seekg (this=0x7fffffffd280, __off=0, __dir=std::_S_end) at /build/gcc/src/gcc-build/x86_64-pc-linux-gnu/libstdc++-v3/include/bits/istream.tcc:915
#1  0x00000000007aea08 in Falltergeist::Format::Dat::Item::_initialize (this=0x207a1f0) at /home/jsimek/Development/falltergeist/src/Format/Dat/Item.cpp:63
#2  0x00000000007c9194 in Falltergeist::Format::Map::File::_initialize (this=0x207a1f0) at /home/jsimek/Development/falltergeist/src/Format/Map/File.cpp:70
[...]

PR #488, opened 2016-10-20
Fix crash when loading from datadir #486

JanSimek:
I understand this is not what you wanted @isage, but since this issue hasn't been fixed yet I thought having a workaround is better than nothing.

By move semantics you mean using std::unique_ptr or something like that? Also is this the correct way to submit pull requests - from a dedicated branch?

Referencing issue #486.


phobos2077:
@JanSimek, may be late for an answer, but we had been using a "convention" for branches like this `issue/123-my-important-issue`.

PR #490, opened 2016-10-26
Implement Handler For Opcode 8138 (mostly)

bludwulf:
Handler for opcode 8138 mostly done, but it doesn't take into account container recursion yet. (like if you have a bag in your inventory with bottlecaps inside of it).

### DIFF
--- a/src/VM/Handler/Opcode8138Handler.cpp
+++ b/src/VM/Handler/Opcode8138Handler.cpp
@@ -25,6 +25,8 @@
 // Falltergeist includes
 #include "../../Logger.h"
 #include "../../VM/Script.h"
+#include "../../Game/ContainerItemObject.h"
+#include "../../Game/CritterObject.h"
 
 // Third party includes
 
@@ -40,9 +42,40 @@ namespace Falltergeist
 
             void Opcode8138::_run()
             {
+                //TODO: search ContainerItemObject and CritterObject inventories recursively if it has container itemobjects in it
+                //for now recursive inventories are not yet implemented since ItemObject has no inventory vector
+                
                 Logger::debug("SCRIPT") << "[8138] [=] int item_caps_total(void* obj)" << std::endl;
-                _script->dataStack()->popObject();
-                _script->dataStack()->push(0);
+                auto object = _script->dataStack()->popObject();
+                if (auto container = dynamic_cast<Game::ContainerItemObject*>(object))
+                {
+                    for(auto item: *container->inventory())
+                    {   
+                        if(item->PID() == 519)
+                        {
+                            _script->dataStack()->push(item->amount());
+                            return;
+                        }
+                    }
+                    _script->dataStack()->push(0);
+                }
+                else if (auto critter = dynamic_cast<Game::CritterObject*>(object))
+                {
+                    for(auto item: *critter->inventory())
+                    {   
+                        if(item->PID() == 519)
+                        {
+                            _script->dataStack()->push(item->amount());
+                            return;
+                        }
+                    }
+                    _script->dataStack()->push(0);
+                }
+                else
+                {
+                    _error("item_caps_total: object type has no inventory!");
+                    _script->dataStack()->push(0);
+                }
             }
         }
     }


src/VM/Handler/Opcode8138Handler.cpp
```diff
@@ -40,9 +42,40 @@ namespace Falltergeist
...
+                
                 Logger::debug("SCRIPT") << "[8138] [=] int item_caps_total(void* obj)" << std::endl;
-                _script->dataStack()->popObject();
-                _script->dataStack()->push(0);
+                auto object = _script->dataStack()->popObject();
+                if (auto container = dynamic_cast<Game::ContainerItemObject*>(object))
+                {
+                    for(auto item: *container->inventory())
+                    {   
+                        if(item->PID() == 519)
```
bludwulf: that's bottlecap Prototype ID according to the fallout wiki. I wasn't sure how to check what item it was. It's definitely not good because fallout 1 bottlecap prototype id is different.

Maybe change it to check by item name, or is there some better way to identify items?

phobos2077: It's bad practice to hard-code magic numbers. Replace with ENUM reference (as temporary solution), or implement properly (like checking if it's an actual container).

bludwulf: the only other ways i can think of doing it is an enum full of all the item pid's and reference it from this enum, but bottlecaps pid would still be hardcoded, just in an enum

The other way would be searching through the inventory vector of items by name, which would probably be pretty slow if it needs to do that often and the item name would be hardcoded.

How would i check how the original engine did it?

phobos2077: You can't check by name, because name is loaded from translation files, it's different depending on game installation... 

I just checked the engine, you just iterate over inventory contents and compare PID with enum value of PID_BOTTLE_CAPS and also use recursion if item subtype is container.

bludwulf: Should it be PID_BOTTLE_CAPS or PID_REAL_BOTTLE_CAPS? Also, /src/Format/Enum.h should be where this enum is defined, right?
bottlecaps

src/VM/Handler/Opcode8138Handler.cpp
```diff
@@ -36,47 +39,38 @@ namespace Falltergeist
...
                 {
                     _error("item_caps_total: object type has no inventory!");
                     _script->dataStack()->push(0);
                 }
             }
+
+            unsigned int Opcode8138::countBottlecaps(std::vector<Game::ItemObject*>* inventory, unsigned int bottlecaps)
+            {
+                unsigned int totalBottlecaps = bottlecaps;
+                for(auto item: *inventory)
```
phobos2077: Space after `for` and `if`.

src/VM/Handler/Opcode8138Handler.cpp
```diff
@@ -36,47 +39,38 @@ namespace Falltergeist
...
+                {
+                    if(auto containerItem = dynamic_cast<Game::ContainerItemObject*>(item))
+                        totalBottlecaps += countBottlecaps(containerItem->inventory(), totalBottlecaps);
+                    else if(item->PID() == (int)ITEM_PID::PID_BOTTLE_CAPS)
+                        totalBottlecaps += item->amount();
```
phobos2077: Put conditional blocks in brackets.

src/VM/Handler/Opcode8138Handler.cpp
```diff
@@ -21,12 +21,15 @@
...
 // C++ standard includes
+#include <vector>
 
 // Falltergeist includes
 #include "../../Logger.h"
 #include "../../VM/Script.h"
 #include "../../Game/ContainerItemObject.h"
 #include "../../Game/CritterObject.h"
+#include "../../Game/ItemObject.h"
+#include "../../Format/Enums.h"
```
phobos2077: Not sure if it's appropriate to use File Format-related enums in our engine code. Basically everything under `Format` namespace should be used solely when loading files (via `ResourceManager`).

src/VM/Handler/Opcode8138Handler.cpp
```diff
@@ -36,47 +39,38 @@ namespace Falltergeist
...
-                }
+                    _script->dataStack()->push(countBottlecaps(critter->inventory(), 0));
                 else
                 {
                     _error("item_caps_total: object type has no inventory!");
                     _script->dataStack()->push(0);
                 }
             }
+
+            unsigned int Opcode8138::countBottlecaps(std::vector<Game::ItemObject*>* inventory, unsigned int bottlecaps)
```
phobos2077: I think the intention of this code will be clearer if you use reference instead of a pointer for `inventory` argument. Here it's not clear if inventory can be null or not.

PR #495, opened 2017-03-19
Try update travis

phobos2077:
**DO NOT MERGE**. Just for testing travis configs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
 language: cpp
 sudo: required
 osx_image: xcode7.3
-dist: precise
+dist: trusty
 
 script:
   - cmake .
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@
 # It must be set before project
 set(CMAKE_BUILD_TYPE "Debug" CACHE STRING "Choose the type of build (by default Debug)")
 
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.1)
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_SOURCE_DIR}/cmake/modules)
 
 project(falltergeist)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,12 +77,9 @@ endif()
 
 file(GLOB_RECURSE SOURCES  src/*.cpp)
 
-if(CMAKE_SYSTEM_NAME MATCHES "Windows")
-  if(MINGW)
-	add_definitions(-std=c++14 -Wall)
-  endif()
-else()
-    add_definitions (-std=c++14 -Wall)
+set(CMAKE_CXX_STANDARD 14)
+if(NOT CMAKE_SYSTEM_NAME MATCHES "Windows" OR MINGW)
+	add_definitions(-Wall)
 endif()
 
 if(CMAKE_SYSTEM_NAME MATCHES "Windows")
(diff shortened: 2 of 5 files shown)


janisozaur:
you will need `set(CMAKE_CXX_STANDARD 11)`

janisozaur:
See https://github.com/falltergeist/falltergeist/pull/496

phobos2077:
@janisozaur : Setting CMAKE_CXX_STANDARD doesn't fix it for some reason. Can you help me out on this?

janisozaur:
@phobos2077 please cherry pick https://github.com/falltergeist/falltergeist/pull/496/commits/23cfeb56c78d070a18048b20d546884e0eee4150 as well

janisozaur:
I just checked, trusty ships with cmake 2.8, so it won't work, sorry.

http://packages.ubuntu.com/trusty/cmake

I generally don't deal with outdated software, so you might want to use something else or do it manually until you can use something that's more current.

In the project I mainly contribute to, OpenRCT2, I changed all of our workers to use docker as it allowed me to use Ubuntu version that is not horribly outdated, just normal Debian-outdated. It was fairly straight-forward, see https://github.com/OpenRCT2/OpenRCT2/pull/4881

phobos2077:
@janisozaur , yes I was just thinking about switching to docker container. I hesitated because we have sh install scripts which use sudo which is not available in those containers.

janisozaur:
What do you use sudo for? Ideally, you should be able to just do `make DESTDIR=/usr install` or `make package`, both served by CPack, a component of CMake. This is how our semi-official package does it: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=openrct2-git

Do note that you can still use your sudo/sh outside the docker container, which you can use to compile the code only.

isage:
PrecompiledHeader.cmake doesn't support CXX_STANDARD and also have problems with -g (it always builds with -gnone, even in debug builds). Either revert it or fix it. And never ever again merge branches with build failures.

For fresh cmake use `sudo add-apt-repository ppa:george-edison55/cmake-3.x -y`
@janisozaur we are using sudo for ppa's and deps. I see no way of installing fresh deps on docker infra.

While at it, you may also add 
```
matrix:
  exclude:
  - os: osx
    compiler: gcc
```
to travis.yml

phobos2077:
@isage , well the scripts are fine, they all work with recent CMake. The problem is travis config using old software for builds. This is why I'm trying to update it so we don't struggle with these silly "ancient sofware" issues in future.

janisozaur:
@isage you are not using it properly then. You shouldn't build a whole image when you start a travis job. The image should already exist and be ready for pulling – that's what we do in OpenRCT2. The images are being built on docker hub automatically (with all kinds of root access you may need), then only get pulled to travis when performing a build for a commit.

That way you also save quite a bit of time, I think converting to docker images saved some 30-60s per build

phobos2077:
@janisozaur : Faster builds sound cool. If only someone could configure them. Certainly not me, these linux commands look like magic to me :)

isage:
@janisozaur you keep mentioning OpenRTC2, yet you are running docker inside of legacy travis builder https://github.com/OpenRCT2/OpenRCT2/blob/develop/.travis.yml
We don't really need that and that doesn't speed up builds at all.

janisozaur:
@isage yes, I am using docker on the legacy infrastructure. This is because there is not enough control over the "native" container infra on travis.

It speeds up the overall build time, because we don't waste travis' time on building the _system_, but our own code. Travis uses pre-made images prepared by hub.docker.com

At least in our case, when we still were tied to 32-bit-only code, installing all libraries on native system cause rebuild of initrd… This is just wasted time.

PR #496, opened 2017-03-19
Set C++ standard in CMakeLists.txt

janisozaur:


janisozaur:
This will only work with https://github.com/falltergeist/falltergeist/pull/495

CMakeLists.txt
```diff
@@ -77,12 +77,13 @@ endif()
 
 file(GLOB_RECURSE SOURCES  src/*.cpp)
 
+set(CMAKE_CXX_STANDARD 14)
 if(CMAKE_SYSTEM_NAME MATCHES "Windows")
```
Mixaill: Can this be simplified just to 
```cmake
if(NOT MSVC)
    add_definitions(-Wall)
endif()
```

 ?

phobos2077:
Overruled by #497 .

PR #497, opened 2017-03-20
Precompiled header fix for linux

phobos2077:
Fixed precompiled header compilation on linux with gcc;
Updated travis build to newer machine;
Replaced -std=  flags with CMAKE_CXX_STANDARD variable;
Updated minimum CMake version to 3.1

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,19 +77,24 @@ endif()
 
 file(GLOB_RECURSE SOURCES  src/*.cpp)
 
-set(CMAKE_CXX_STANDARD 14)
-if(NOT MSVC)
+if(MSVC)
+	set(_additional_includes "package/windows.rc" pchheader.cpp)
+endif()
+
+add_executable(falltergeist main.cpp ${SOURCES} ${_additional_includes})
+
+set_target_properties(falltergeist PROPERTIES
+	CXX_STANDARD 14
+	CXX_STANDARD_REQUIRED YES
+	CXX_EXTENSIONS NO
+)
+
+if(MSVC)
+	include(cmake/modules/PrecompiledHeader.cmake)
+	add_precompiled_header(falltergeist pchheader.h FORCEINCLUDE SOURCE_CXX pchheader.cpp)
+else()
 	add_definitions(-Wall)
 endif()
-
-if(CMAKE_SYSTEM_NAME MATCHES "Windows")
-	add_executable(falltergeist main.cpp ${SOURCES} "package/windows.rc" pchheader.cpp)
-else()
-	add_executable(falltergeist main.cpp ${SOURCES} pchheader.cpp)
-endif()
-
-include(cmake/modules/PrecompiledHeader.cmake)
-add_precompiled_header(falltergeist pchheader.h FORCEINCLUDE SOURCE_CXX pchheader.cpp)
 
 if (CONAN_LIBS)
 	target_link_libraries(falltergeist ${CONAN_LIBS})
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,7 +78,7 @@ endif()
 file(GLOB_RECURSE SOURCES  src/*.cpp)
 
 set(CMAKE_CXX_STANDARD 14)
-if(NOT CMAKE_SYSTEM_NAME MATCHES "Windows" OR MINGW)
+if(NOT MSVC)
 	add_definitions(-Wall)
 endif()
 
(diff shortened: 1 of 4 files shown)


phobos2077:
Now the build is successful, but I don't see dramatic compile speed increase for GCC linux where this header seemed to work (from the build log). If anyone who uses linux can check and tell if compile times decreased or not, I would appreciate it.

In my case on Windows with Visual Studio 2017, compilation is about 4 times faster with these changes.

janisozaur:
I can try checking it later today.

cmake/modules/PrecompiledHeader.cmake
```diff
@@ -171,19 +171,24 @@ function(add_precompiled_header _target _input)
     set(_pch_flags_file "${_pch_binary_dir}/compile_flags.rsp")
     export_all_flags("${_pch_flags_file}")
     set(_compiler_FLAGS "@${_pch_flags_file}")
+    if(CMAKE_CXX_STANDARD EQUAL 11)
+        set(_cpp_standard_FLAG "-std=c++11")
```
isage: without above change you'll get -std=c++11 for pch and -std=gnu++11 for everything else

cmake/modules/PrecompiledHeader.cmake
```diff
@@ -171,19 +171,24 @@ function(add_precompiled_header _target _input)
...
-      COMMAND "${CMAKE_CXX_COMPILER}" ${_compiler_FLAGS} -x c++-header -o "${_output_cxx}" "${_pchfile}"
+      COMMAND "${CMAKE_CXX_COMPILER}" ${_compiler_FLAGS} ${_cpp_standard_FLAG} -x c++-header -o "${_output_cxx}" "${_pchfile}"
       DEPENDS "${_pchfile}" "${_pch_flags_file}"
       COMMENT "Precompiling ${_name} for ${_target} (C++)")
     add_custom_command(
       OUTPUT "${_output_c}"
-      COMMAND "${CMAKE_C_COMPILER}" ${_compiler_FLAGS} -x c-header -o "${_output_c}" "${_pchfile}"
+      COMMAND "${CMAKE_C_COMPILER}" ${_compiler_FLAGS} ${_cpp_standard_FLAG} -x c-header -o "${_output_c}" "${_pchfile}"
       DEPENDS "${_pchfile}" "${_pch_flags_file}"
       COMMENT "Precompiling ${_name} for ${_target} (C)")
```
isage: Please also fix -g for debug builds.

phobos2077: It's not getting passed via _compiler_FLAGS ?

CMakeLists.txt
```diff
@@ -77,12 +77,9 @@ endif()
 
 file(GLOB_RECURSE SOURCES  src/*.cpp)
 
-if(CMAKE_SYSTEM_NAME MATCHES "Windows")
-  if(MINGW)
-	add_definitions(-std=c++14 -Wall)
-  endif()
-else()
-    add_definitions (-std=c++14 -Wall)
+set(CMAKE_CXX_STANDARD 14)
```
phobos2077: This way my solution for PrecompiledHeader.cmake will not work. Oh well...

cmake/modules/PrecompiledHeader.cmake
phobos2077: What do you mean "without above change"?

cmake/modules/PrecompiledHeader.cmake
isage: Looks like it's not. My compiler complains that pch was built with -gnone, but used with -gdwarf-2

cmake/modules/PrecompiledHeader.cmake
isage: Without CXX_EXTENSIONS NO cmake uses `-std=gnu++*` instead of `-std=c++*`

phobos2077: Oh, I see. Now need to figure out how to check value of CXX_STANDARD or just simply hard-code the correct flag.

isage: Look how PrecompiledHeader.cmake gets other target properties

cmake/modules/PrecompiledHeader.cmake
phobos2077: I wish travis compiler would also complain, instead of silent ignore :)

cmake/modules/PrecompiledHeader.cmake
phobos2077: I looked but didn't understand much. Most of this CMake code is way above my level of understanding :)

phobos2077:
Since I don't have enough knowledge to fix the script for GCC build myself, I think the best option is just add condition for using PCH with MSVC only, until some linux user can help.

PR #513, opened 2017-04-02
Hexagon and Sprite Changes

Zervox:
Changes to Sprite renderScaled and Rendercropped

Changes to Hexagon to use std::array instead of std::vector
Changes to Hexagon grid initialization of Hexagon neighbours and Pathfind

### DIFF
--- a/src/Graphics/Sprite.cpp
+++ b/src/Graphics/Sprite.cpp
@@ -164,14 +164,14 @@ void Sprite::renderScaled(int x, int y, unsigned int width, unsigned int height,
 
     GL_CHECK(glBindBuffer(GL_ARRAY_BUFFER, Game::getInstance()->renderer()->getVVBO()));
 
-    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, 4 * sizeof(glm::vec2), &vertices, GL_DYNAMIC_DRAW));
+    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, sizeof(vertices) * sizeof(glm::vec2), &vertices[0], GL_DYNAMIC_DRAW));
 
     GL_CHECK(glVertexAttribPointer(_attribPos, 2, GL_FLOAT, GL_FALSE, 0, (void*)0 ));
 
 
     GL_CHECK(glBindBuffer(GL_ARRAY_BUFFER, Game::getInstance()->renderer()->getTVBO()));
 
-    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, 4 * sizeof(glm::vec2), &UV, GL_DYNAMIC_DRAW));
+    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, sizeof(UV) * sizeof(glm::vec2), &UV[0], GL_DYNAMIC_DRAW));
 
     GL_CHECK(glVertexAttribPointer(_attribTex, 2, GL_FLOAT, GL_FALSE, 0, (void*)0 ));
 
@@ -287,14 +287,14 @@ void Sprite::renderCropped(int x, int y, int dx, int dy, unsigned int width, uns
 
     GL_CHECK(glBindBuffer(GL_ARRAY_BUFFER, Game::getInstance()->renderer()->getVVBO()));
 
-    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, 4 * sizeof(glm::vec2), &vertices, GL_DYNAMIC_DRAW));
+	GL_CHECK(glBufferData(GL_ARRAY_BUFFER, sizeof(vertices) * sizeof(glm::vec2), &vertices[0], GL_DYNAMIC_DRAW));
 
     GL_CHECK(glVertexAttribPointer(_attribPos, 2, GL_FLOAT, GL_FALSE, 0, (void*)0 ));
 
 
     GL_CHECK(glBindBuffer(GL_ARRAY_BUFFER, Game::getInstance()->renderer()->getTVBO()));
 
-    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, 4 * sizeof(glm::vec2), &UV, GL_DYNAMIC_DRAW));
+	GL_CHECK(glBufferData(GL_ARRAY_BUFFER, sizeof(UV) * sizeof(glm::vec2), &UV[0], GL_DYNAMIC_DRAW));
 
     GL_CHECK(glVertexAttribPointer(_attribTex, 2, GL_FLOAT, GL_FALSE, 0, (void*)0 ));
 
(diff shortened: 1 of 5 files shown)


src/Graphics/Sprite.h
```diff
@@ -72,6 +69,8 @@ class Sprite
 
     GLint _attribPos;
     GLint _attribTex;
+	Texture* _texture;
+	Graphics::TransFlags::Trans _trans = Graphics::TransFlags::Trans::NONE;
```
phobos2077: We use 4 spaces for indentation, not tabs.

src/Graphics/Sprite.cpp
```diff
@@ -323,14 +287,14 @@ void Sprite::renderCropped(int x, int y, int dx, int dy, unsigned int width, uns
...
-    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, vertices.size() * sizeof(glm::vec2), &vertices[0], GL_DYNAMIC_DRAW));
+    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, 4 * sizeof(glm::vec2), &vertices, GL_DYNAMIC_DRAW));
 
     GL_CHECK(glVertexAttribPointer(_attribPos, 2, GL_FLOAT, GL_FALSE, 0, (void*)0 ));
 
 
     GL_CHECK(glBindBuffer(GL_ARRAY_BUFFER, Game::getInstance()->renderer()->getTVBO()));
 
-    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, UV.size() * sizeof(glm::vec2), &UV[0], GL_DYNAMIC_DRAW));
+    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, 4 * sizeof(glm::vec2), &UV, GL_DYNAMIC_DRAW));
```
phobos2077: `sizeof(UV)` will return correct total size for the whole array.

src/PathFinding/Hexagon.cpp
```diff
@@ -34,26 +34,22 @@ namespace Falltergeist
 
 Hexagon::Hexagon()
 {
+	for (int i = 0; i < _neighbors.size(); i++) {
+		_neighbors[i] = nullptr;
+	}
```
phobos2077: Is this necessary? Can't you do the same with std::array initializer?

src/PathFinding/Hexagon.cpp
```diff
@@ -34,26 +34,22 @@ namespace Falltergeist
...
 }
 
 Hexagon::Hexagon(unsigned int number)
 {
     setNumber(number);
+	for (int i = 0; i < 6; i++) {
+		_neighbors[i] = nullptr;
+	}
```
phobos2077: See above, also it's duplicated code.

src/PathFinding/Hexagon.cpp
```diff
@@ -34,26 +34,22 @@ namespace Falltergeist
...
 }
 
-unsigned int Hexagon::number()
+std::array<Hexagon*, 6> &Hexagon::neighbors()
```
phobos2077: Place reference sign near type, not the function.

src/Graphics/Sprite.h
```diff
@@ -48,16 +48,13 @@ class Sprite
     void render(int x, int y, bool transparency = false, bool light = false, int outline = 0, unsigned int lightValue=0);
     void renderCropped(int x, int y, int dx, int dy, unsigned int width, unsigned int height, bool transparency = false,
                        bool light = false, unsigned int lightValue=0);
-    unsigned int width() const;
-    unsigned int height() const;
-    Size size() const;
+	inline Size size() const {return _texture->size();}
+	inline unsigned int width() const{return _texture->width();}
+	inline unsigned int height() const{return _texture->height();}
```
Zervox: old habit

src/Graphics/Sprite.h
Zervox: sorry,  my project settings currently are default to visual studio which means enter and copy paste will auto tab them

src/PathFinding/Hexagon.cpp
Zervox: old and bad habit.

Zervox:
not sure why it complains about undeclared identifier on the maps, they work fine compiling for me.
never used build systems like travis before.

Zervox:
ok, travis is a bit more sensitive to includers, good to know.

src/Graphics/Sprite.cpp
```diff
@@ -190,14 +164,14 @@ void Sprite::renderScaled(int x, int y, unsigned int width, unsigned int height,
...
-    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, vertices.size() * sizeof(glm::vec2), &vertices[0], GL_DYNAMIC_DRAW));
+    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, sizeof(vertices) * sizeof(glm::vec2), &vertices[0], GL_DYNAMIC_DRAW));
 
     GL_CHECK(glVertexAttribPointer(_attribPos, 2, GL_FLOAT, GL_FALSE, 0, (void*)0 ));
 
 
     GL_CHECK(glBindBuffer(GL_ARRAY_BUFFER, Game::getInstance()->renderer()->getTVBO()));
 
-    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, UV.size() * sizeof(glm::vec2), &UV[0], GL_DYNAMIC_DRAW));
+    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, sizeof(UV) * sizeof(glm::vec2), &UV[0], GL_DYNAMIC_DRAW));
```
phobos2077: sizeof returns the size of variable is bytes, not the number of elements, so this code is wrong.

phobos2077:
Might be more effective to make smaller PRs - each aimed at solving specific issue. Here you have 2 different issues fixed and inlining functions without explanation of why it's better.

src/Graphics/Sprite.cpp
Zervox: and you commented on the change before that .size reported length but seeing as it is not using std::vector anymore .size() is not available.

Zervox:
inline is a suggestion for the compiler, not always does the compiler for each platform know wether or not its worth inlining, inlining these functions is a habit I have to make sure small pieces of code gets inlined(due to suggestion). I've had cases were VS17,15 and 10 does not inline functions which are well suited for inlining, which is why I have a habit of inlining these small functions.

sadly I only have experience with one way interactions of forks, so I am still learning. but thanks I will keep that in mind.

phobos2077:
@Zervox : I still don't see enough justification for making our getters and setters inline and moving them to headers. Can you compile in VS2017 in Release mode and confirm that those getters are not inlined by examining assembly code?

Otherwise it's just a lot of unnecessary changes to code without real benefit.

PR #517, opened 2017-04-04
HexagonGrid and Hexagon Changes

Zervox:


phobos2077:
The same issues as other PRs. Also no description...

JanSimek:
@phobos2077 maybe you should write a [clang-format](https://clang.llvm.org/docs/ClangFormat.html) or [uncrustify](http://uncrustify.sourceforge.net/) configuration, so that the code can be automatically formatted to falltergeist style

phobos2077:
@JanSimek that's a good idea. But I think stuff like tabs and brackets style are obvious when you just study some project code. 95% of our code has consistent style, more or less, so the code itself should be pretty self explanatory. When people ignore existing style I think they won't use any automation tool either.

Hm.. I think there is no VS2017 extension for ClangFormat yet. Sad.

eirnym:
@phobos2077 tabs and brackets are not so obvious and I've failed to find any description about your code style (I haven't searched in Pull Requests though as inpropariate place for it). I met various formats and code styles which would conflict with each other, so it's better to describe it. 

There's many ways how to specify code style. The most important is special section in developer documentation if you have any published in the network. The second important place is editor configuration, like [editor config](http://github.com/editorconfig/editorconfig), configurations for several common editors in your project like VS2017, CLion, Vim, Emacs (fill any useful). And finally, the auto-formatting files like mentioned above.

phobos2077:
@eirnym 
https://github.com/falltergeist/falltergeist/wiki/Code-style

eirnym:
@phobos2077 thank you. Please, disregard my comment about developer documentation.

PR #518, opened 2017-04-04
Update to sprite render functions

Zervox:
Changed to render functions

### DIFF
--- a/src/Graphics/Sprite.cpp
+++ b/src/Graphics/Sprite.cpp
@@ -88,18 +88,18 @@ unsigned int Sprite::height() const
 // render, optionally scaled
 void Sprite::renderScaled(int x, int y, unsigned int width, unsigned int height, bool transparency, bool light, int outline, unsigned int lightValue)
 {
-	glm::vec2 vertices[4] = {
-		glm::vec2((float)x, (float)y),
-		glm::vec2((float)x, (float)(y + height)),
-		glm::vec2((float)(x + width), (float)y),
-		glm::vec2((float)(x + width), (float)(y + height))
-	};
-	glm::vec2 UV[4] = {
-		glm::vec2(0.0, 0.0),
-		glm::vec2(0.0, (float)_texture->height() / (float)_texture->textureHeight()),
-		glm::vec2((float)_texture->width() / (float)_texture->textureWidth(), 0.0),
-		glm::vec2((float)_texture->width() / (float)_texture->textureWidth(), (float)_texture->height() / (float)_texture->textureHeight())
-	};
+    glm::vec2 vertices[4] = {
+        glm::vec2((float)x, (float)y),
+        glm::vec2((float)x, (float)(y + height)),
+        glm::vec2((float)(x + width), (float)y),
+        glm::vec2((float)(x + width), (float)(y + height))
+    };
+    glm::vec2 UV[4] = {
+        glm::vec2(0.0, 0.0),
+        glm::vec2(0.0, (float)_texture->height() / (float)_texture->textureHeight()),
+        glm::vec2((float)_texture->width() / (float)_texture->textureWidth(), 0.0),
+        glm::vec2((float)_texture->width() / (float)_texture->textureWidth(), (float)_texture->height() / (float)_texture->textureHeight())
+    };
 
     x--;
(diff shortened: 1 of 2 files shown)


src/Graphics/Sprite.cpp
```diff
@@ -190,14 +164,14 @@ void Sprite::renderScaled(int x, int y, unsigned int width, unsigned int height,
 
     GL_CHECK(glBindBuffer(GL_ARRAY_BUFFER, Game::getInstance()->renderer()->getVVBO()));
 
-    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, vertices.size() * sizeof(glm::vec2), &vertices[0], GL_DYNAMIC_DRAW));
+    GL_CHECK(glBufferData(GL_ARRAY_BUFFER, 4 * sizeof(glm::vec2), &vertices[0], GL_DYNAMIC_DRAW));
```
phobos2077: You can just `sizeof(vertices)`  here.

src/Graphics/Sprite.cpp
```diff
@@ -223,28 +197,18 @@ void Sprite::render(int x, int y, bool transparency, bool light, int outline, un
...
+		glm::vec2((float)x, (float)(y + height)),
+		glm::vec2((float)(x + width), (float)y),
+		glm::vec2((float)(x + width), (float)(y + height))
+	};
+    glm::vec2 UV[4] = {
+		glm::vec2((float)dx / (float)_texture->textureWidth(), (float)dy / (float)_texture->textureHeight()),
+		glm::vec2((float)dx / (float)_texture->textureWidth(), (float)(dy + height) / (float)_texture->textureHeight()),
+		glm::vec2((float)(dx + width) / (float)_texture->textureWidth(), (float)dy / (float)_texture->textureHeight()),
+		glm::vec2((float)(dx + width) / (float)_texture->textureWidth(), (float)(dy + height) / (float)_texture->textureHeight())
+	};
```
phobos2077: Please follow the style guide. We don't use tabs. I already mentioned this before.

phobos2077:
This PR is trying to "slip through" another change (inline setters/getters) without any kind of description. I don't understand why we should do that, so someone need to resource this topic before this can be merged.

phobos2077:
Ok I've investigated a little bit about inline functions. Here's my conclusion:

- Let's **never use `inline` keyword** in our code. It's just another keyword to use without real benefit (compilers know when to inline).
- Declaring functions `inline` inside class declaration is completely redundant by itself, because by C++ standard all functions that defined in the header are inline implicitly.
- It may be fine to move implementation of trivial getters and setters into the header (without the `inline`), but there is no reason to do so for existing code. No benefit.

janisozaur:
> Declaring functions `inline` inside class declaration is completely redundant by itself, because **by C++ standard all functions that defined in the header are inline implicitly**.

Can you provide exact sources for that?

phobos2077:
http://en.cppreference.com/w/cpp/language/inline

janisozaur:
What still baffles me, even after reading the source you provided, is why do you say "all functions that defined in the header are inline implicitly". I can find no reference to that, but perhaps I'm not reading it well enough.

I don't see how being in a "header file" or in class declaration is relevant here at all, really.

phobos2077:
Probably I didn't rephrased it very well. Here is the quote:

> A function defined entirely **inside a class/struct/union definition**, whether it's a member function or a non-member friend function, is implicitly an inline function. 

I've found the same statement in other sources as well.

Zervox:
if one is to argue for redundancy the inline gets pruned by the compiler(there still are cases where it does not inline without inline, even if it is suitable). the compiler is not a magical know it all tool unless you actually performance profile the project.

the statements that you-ve looked up regarding all functions inside header is implicitly inlined is false and is proven many times over, this depends on the compiler and each compiler might inline different functions where another does not. 90% of cases each compiler does inline functions, but sometimes it does not and it could very well be a really good function to inline.

yet there is duplicates, like the Parser of Ini and Txt, Point and Size(does the exact same thing even casts to one another) but Size has a useless check inside its setters which only gets encountered if one has fed the wrong values, in which case it just sets it to 0 so you still see nothing on screen.

PR #520, opened 2017-04-04
Msgchanges

Zervox:
changes to Msg::Message and Msg::File prevent duplicate entries when looking for description of items

### DIFF
--- a/src/Format/Msg/File.h
+++ b/src/Format/Msg/File.h
@@ -50,11 +50,11 @@ public:
     File(std::ifstream* stream);
     ~File();
 
-    std::vector<Message*>* messages();
     Message* message(unsigned int number);
 
-protected:
+
     std::map<uint32_t,Message*> _messages;
+protected:
     virtual void _initialize();
 
 };
--- a/src/Format/Msg/File.h
+++ b/src/Format/Msg/File.h
@@ -26,7 +26,7 @@
 #define FALLTERGEIST_FORMAT_MSG_FILE_H
 
 // C++ standard includes
-#include <vector>
+#include <map>
 
 // Falltergeist includes
 #include "../../Format/Dat/Item.h"
@@ -54,7 +54,7 @@ public:
     Message* message(unsigned int number);
 
 protected:
-    std::vector<Message*> _messages;
+    std::map<uint32_t,Message*> _messages;
     virtual void _initialize();
 
 };
--- a/src/Format/Msg/Message.h
+++ b/src/Format/Msg/Message.h
@@ -44,24 +44,12 @@ class Message
 
 public:
     Message();
+	Message(std::string &s, std::string &t);
     ~Message();
 
-    void setNumber(unsigned int number);
-    unsigned int number();
-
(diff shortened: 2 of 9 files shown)


src/Format/Msg/Message.h
```diff
@@ -44,24 +44,12 @@ class Message
...
-    std::string text();
-    std::string* textPointer();
-
-    operator std::string();
-
-protected:
-    unsigned int _number = 0;
     std::string _sound;
     std::string _text;
+	std::string _desc;
```
phobos2077: Are these public properties? Issues:
- Why do you remove getters? 
- Public properties should not start with undescore.
- `_desc` property makes no sense. It duplicates `_text`.

src/Format/Msg/Message.h
```diff
@@ -44,24 +44,12 @@ class Message
 
 public:
     Message();
+	Message(std::string &s, std::string &t);
```
phobos2077: Why not `const std::string&`? Such interface make the intent of those arguments obscure. In all such cases either use `const std::string&` (read-only use or copy) or `std::string&&` (move). Here const reference is obvious choice.

src/Game/ObjectFactory.cpp
```diff
@@ -135,8 +135,9 @@ Object* ObjectFactory::createObject(unsigned int PID)
             auto msg = ResourceManager::getInstance()->msgFileType("text/english/game/pro_item.msg");
             try
             {
-                object->setName(msg->message(proto->messageId())->text());
-                object->setDescription(msg->message(proto->messageId() + 1)->text());
+				auto o = msg->message(proto->messageId());
```
phobos2077: Don't use tabs.

src/Format/Msg/File.h
```diff
@@ -50,11 +50,11 @@ class File : public Dat::Item
     File(std::ifstream* stream);
     ~File();
 
-    std::vector<Message*>* messages();
     Message* message(unsigned int number);
 
+
+    std::map<uint32_t,Message*> _messages;
```
phobos2077: Why not `unordered_map`? Also this property should not be public.

phobos2077:
There aren't any actual duplicates in vanilla `msg` files, but the check for this is nice to have for robustness.

Zervox:
actually there is, when it _initializes it checks for files and their descriptor data, the description data gets put into MSG+1 value, meaning you have to look up two memory locations for the same data(and the only area which is checked at runtime is when getting description and name of things you are looking for.

src/Game/ObjectFactory.cpp
Zervox: I am not using tabs, I am using enter, VS2017 does it whenever I reinsert removed lines.

phobos2077:
I think you don't understand how MSG files work. Each message ID is separate, there is no inherent semantic connection between them. The fact that names of objects and their description go together - is just how developers organized their messages. But they should be read individually and each message has only 3 data fields - number, sound and text. Nothing more.

PR #522, opened 2017-04-06
Fix the Player Panel offset

Zervox:
Fixes the player panel offset.

the same should be done for doors(not sure if this actually counts for all scenery objects or only door objects, this also explains the difficulty selecting the door to Slim to get to the mine)

for game objects would probably to override _generateUi as it seems input is also offset by the same amount(need to find out if its only doors or all scenery objects that are offset

### DIFF
--- a/src/UI/PlayerPanel.cpp
+++ b/src/UI/PlayerPanel.cpp
@@ -62,8 +62,8 @@ PlayerPanel::PlayerPanel() : UI::Base()
     _background = std::make_shared<Image>("art/intrface/iface.frm");
     _ui.push_back(_background);
 
-    setX((renderer->width() - 640) / 2);
-    setY(renderer->height() - _background->height());
+    setX((renderer->width() - 640) / 2-1);
+    setY(renderer->height() - _background->height()+1);
 
     _background->setPosition(this->position());
 


phobos2077:
Strange. +1 and -1 pixels look like some magic number...

Zervox:
well considering that the iframe is 642 width and 101 height seems like magic all around.

phobos2077:
We need a proper fix for this. I'm sure there is a simple explanation. This is just workaround that probably hides the problem that will bite us in the arse later :)

Zervox:
doubt it, as the 
x value is -1 not 0 with my fix
y value is 380 instead of 379

Zervox:
these magic numbers seems to come from Frm file, currently working on a branch related to this(seems to be something wrong in the rgba function amongst other things)

phobos2077:
@Zervox , just try not to break everything while you're at it :)
There shouldn't be any "magic" with FRMs as far as I know. The only exception were the moving animations (hence the "dirty hack" I tried to make). That's because FRMs only support 1 action frame in their format, but running/walking is quite complicated and has more than 1 "action frame" when you need to switch to the next hex.

Simply displaying UI panel shouldn't have any issues AFAIK. And I don't remember these bugs a year and a half ago when I last worked on the project.

Zervox:
I know the offset to correct PlayerPanel left side edge in the frm(doesn't ruin anything else) the y however I am not entirely sure about after I tested the player panel it had dead pixel on the left(it was part of the UI but it had no data so SDL renders it as seethrough. the same is true for bottom of playerpanel

PR #523, opened 2017-04-06
Seems like the && is an error

Zervox:
objects can't be Wall and Scenery at the same time it should return false instantly, this makes check faster if it's wall, else it will check both. instead of checking both every time

### DIFF
--- a/src/Game/Object.cpp
+++ b/src/Game/Object.cpp
@@ -336,7 +336,7 @@ void Object::render()
 bool Object::_isIntersectsWithEgg()
 {
     //only walls and scenery are affected by egg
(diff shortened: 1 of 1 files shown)


Zervox:
Could be the offset of Frm loading is wrong?

phobos2077:
Have you even tested this change? 
Here's two tables of possible results of this condition. First is how it was, second how you did it. The number is how many checks is made and true/false is the condition result - true means we don't intersect with egg.

```
&&
WALL	 1	false
SCENERY  2  false
OTHER    2  true

||
WALL	 2	true -> BUG
SCENERY  1  true -> BUG
OTHER    2  true
```

As you see, your change makes condition essentially `if (true)` ...

PR #524, opened 2017-04-06
Fix Animation Offset of animated objects

Zervox:
Fixes Animation and Selection offset of animated objects


phobos2077:
This doesn't look correct at all. Have you confirmed the original implementation that it uses the same technique?

Zervox:
xOffset += frm->offsetX(direction, f);
yOffset += frm->offsetY(direction, f);
tbh it seems to be it's not offset enough

phobos2077:
Yeah it was me, but that hack isn't merged yet :)

PR #553, opened 2018-03-11
Pull that cleans up #includes (mostly)

adamkewley:
I'm currently trying to familiarize myself with the codebase so that I can make some meaningful changes. I love fallout 2 and this project is the closest I've seen to an open-source implementation of the engine - very impressive.

In order to familiarize with the codebase, I first wanted to understand the file dependencies (i.e. the general project structure). However, my IDE (clion) found quite a few unused dependencies and several cases where headers couldn't compile if they were not loaded before certain libraries (e.g. some files were implicitly relying on the .cpp loading `<vector>` for them).

As a first exercise with the codebase, I manually started refactoring things that could be simplified (earlier commits - nothing big) but then realized I was spending loads of time manually cleaning up `#includes`. Because of that, I wrote a script that reorganized includes with the following priority:

- (if applicable) related header (that is, the header for a cpp file)
- Project includes (i.e. headers in falltergeist)
- External includes (i.e. anything beginning with `<`)

Loading files in this way ensures that the header won't accidentally rely on the implementation loading something beforehand. It also ensures that the falltergeist headers don't accidentally rely on a 3rd-party lib being loaded beforehand. 

Wherever the build broke, I added/rearranged the dependencies accordingly (it only broke in 3 places, and it was for a `<vector>` and `<map>`). After each rearrangement was complete, I used clion to remove all unused dependencies. In total, around 300 includes have been removed from this commit (old total: ~2950). 

I have tried to put each cleanup step (usually, a folder in the codebase) into a separate commit. However, **not all commits are `#include` cleanups**: the first 2/3 also remove the `Singleton` class and do some renaming. The renaming / code rearrangements are reverted in the final commit (I couldn't think of a clean way of hiding that in a rebase) but the singleton has still been replaced with `static` members instead (the `Singleton` implementation, imho, is more complicated than the problem it's trying to solve and `friend` classes tend to be iffy with tools). 

I know this is many changes for a single set of commits, but most of them are just boilerplate `#include` changes: hopefully, they can be scanned through quite quickly. 

Note: The project compiles + runs fine with these changes: I believe they are all non-destructive (apart from the Singleton change)


janisozaur:
@adamkewley See https://github.com/falltergeist/falltergeist/pull/554

janisozaur:
(Note: I'm just a drive-by contributor) It'd also be nice if you cleaned up your commit messages subjects a bit too:

https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

Otherwise they end up cluttering the whole log when browsing through history with `--oneline` option.

adamkewley:
Closing this because it's screwed up the Travis build (probably because my ordering made the usage optimizer remove too many headers - will need to re-order it differently)

adamkewley:
Thanks @janisozaur, it's my first bigger pull req on github. I wasn't too sure whether the maintainers prefer to squash the commits later or whether they prefer them to be a bit fragmented. I'll redo this pull request with a better model and commit messages (I'll be faster next time ;))

PR #554, opened 2018-03-11
Add -k option to make for Travis

janisozaur:
In cases where compilation fails on a foreign platform, it's useful to
find all the errors at once rather than having to submit a fix for each
discovered error and wait for next one. Make's `-k` option makes it
continue in face of errors until it cannot proceed further, allowing to
surface more errors in one go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ dist: trusty
 
 script:
   - cmake .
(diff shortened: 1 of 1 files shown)


alexeevdv:
Thanks. This can be useful indeed

PR #557, opened 2018-05-28
WIP: VM interface

alexeevdv:


Zervox:
This is very similar to what I did when I rewrote it in my local copy to utilize function pointers instead of separate objects for every opcode handler(this way I could maintain only a single header for all the objects while still keeping it small(easier to scroll singular functions compared to objects) while minimizing duplicate includes etc.

FakelsHub:
Какие-то подвижки в проекте.
Не забудьте сделать поддержку sfall функций(опкодов) :)

PR #558, opened 2018-05-28
automatic code formating

alexeevdv:


phobos2077:
Why would you use that strange code style? :)
Why not just take one of the popular code styles or stick to the one that was already in place (I think I've even described it in the wiki section, so you might want to update that now and provide a config for the tool you have used).

alexeevdv:
It is default formatting in CLion. I dont care that much about formatting currently. It just makes my life easier

PR #559, opened 2018-08-13
Accessing inventory crash fix #531

### LINKED ISSUE #531: Segfault when trying to open inventory after acquiring new item via dialogue
### Description
A Segfault when trying to open inventory after acquiring new item via dialogue.
`InventoryItem.cpp:137` (render function)

### System
OS: Ubuntu 17.04 x64
Compiler: GCC

### Steps to reproduce
1. Start the game on arvillag map.
2. Talk to the Elder.
3. Be obedient and she will give you money and a flask. 
4. Close dialog.
5. Try opening inventory.

667bdrm:
Patch by @JanSimek

### DIFF
--- a/src/VM/Handler/Opcode8116Handler.cpp
+++ b/src/VM/Handler/Opcode8116Handler.cpp
@@ -27,6 +27,7 @@
 #include "../../Game/CritterObject.h"
 #include "../../Game/Game.h"
 #include "../../Game/Object.h"
+#include "../../Game/ObjectFactory.h"
 #include "../../Logger.h"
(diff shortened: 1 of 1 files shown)


JanSimek:
This is a just a workaround to explain why it crashes. I am afraid it is not a proper fix.

667bdrm:
On my side it fixed crash on accessing inventory if artemple.map specified as starting in the config

JanSimek:
Oh yes, It does prevent the crash. I am just saying it is a workaround and not a proper fix for the bug.

PR #561, opened 2018-08-29
Add icon to .desktop file and install icon to pixmaps on Linux

aleksvor:
This makes packaging for Linux cleaner. Currently desktop shortcut installs without icon and looks ugly after installation.


alexeevdv:
Thanks!

Issue #562, opened 2018-10-09
Add milestones or state of the re-implementation


nicman23:
Hello, I just found this project and it is not clear from README.md or the Wiki in what state the engine is and its playability

While Change logs are fine for anyone familiar with the project, for the end user mean little.

So what is the state of the project?

prokopst:
I have to say that state of the project is very unclear to me as well. Is there a stable team of people behind it? Doesn't seem so. Is it randomly alive and dead? More likely.

What makes me sad is that I would love to contribute to the project but for me C++ is no-go.

Don't get me wrong, I think I understand C/C++ well (at least prior C++14). But that's the problem, I understand it well enough to know that C++ is not the most efficient programming language to use especially when raw pointers are used all over the place with unclear ownership of allocated memory. Hello segmentation faults? #530  #531 #538

I do appreciate what all maintainers and contributors achieved, but I think that the lack of any public relations (or how to call it?) and the choice of the programming language + the approach (raw pointers) is a blocker to attract new developers and to make the project alive.

It's all my opinion, don't lynch me for that.

(I noticed this thread from a discord notification)

adamkewley:
@prokopst , I can understand your frustration, but keep in mind:

- It's a remake of an absolutely massive AAA game
- Which was buggy, even when it was initially developed by a paid team containing some of the best RPG game developers available in the late 90s
- Where the assets are now only available in binary form and need to (e.g.) be reverse-engineered through a virtual machine
- Where, even though it's currently lacking, it's already ~30 kLOC of C++, putting it in "large project" territory
- Commits have been made to this repo over a timespan of 5 years (as far as GitHub shows) with off-and-on, but consistent over the long-term, commits
- ~80 % of what's in the master branch was produced by one (presumably unpaid) developer, @alexeevdv 

As for C++ being a "no go", I don't personally understand what language you'd consider to be a "go". Large game engine projects benefit from:

- Static typing: because refactoring a >5 kLOC codebase without typing is hell
- Strong standard library: because reimplementing lists, trees, algorithms, etc. cleanly is a waste of time
- Powerful language features (lambdas, templating, moving, templating, etc.): because every complex project will end up with codegen sooner or later
- Libraries

[...]

PR #563, opened 2018-10-15
Refactor project into separate libraries

adamkewley:
**PROTOTYPE**: Based on discussion in #562 . 

One of the most valuable things Falltergeist offers is a collection of working open-source parsers for Fallout 1/2's data files. The documentation for those formats is scattered (e.g. [on wikis](http://falloutmods.wikia.com/wiki/Category:Fallout_and_Fallout_2_file_formats), [peoples pet projects](https://github.com/adamkewley/fo2dat)) and not verified. Being able to use those parsers independently of the Falltergeist engine will enable developers to write conversion utilities and write their own game engines.

I had a few hours to spare, so I refactored Falltergeist/develop into:

- Core game engine executable (as previous)
- Data format shared library (new)
- Static library containing classes shared by the above (`falltergeist-common-cpp`)

With this project structure, you should be able to build the separate components in one step (if the IDE supports CMake), rather than needing to maintain two separate repos. With further refactoring, cleanup, and documentation we could publish the IO formats as a library that other developers can use independently of the Falltergeist game engine. This will enable developers to write, for example, command line interfaces for converting between MAP and grid files without needing to duplicate code from the falltergeist executable.

@alexeevdv , if you get the chance, could you check this PR out (or my fork) and see what you think about it. This represents a major change in the project structure, and IDEs might not play ball with it. Please comment when you get the chance.


alexeevdv:
742 changes.... jesus....it will take forever to review this :)

falltergeist/src/Game/Game.cpp
```diff
@@ -32,7 +32,7 @@
 #include "../Event/Dispatcher.h"
 #include "../Event/State.h"
 #include "../Exception.h"
-#include "../Format/Gam/File.h"
+#include <falltergeist/io//Gam/File.h>
```
alexeevdv: does double slash put in these includes for a reason?

adamkewley: That's because I had to manually sed some replacements in. It's a bug :(

adamkewley:
I can try to move them into a separate repo, sure =) The main idea behind keeping it in the mainline was so that you can change the IO code in-branch without needing to pull releases from elsewhere

adamkewley:
@alexeevdv I'd close this PR for now. Now that I know what you want I'll get to work on developing a separate project.

alexeevdv:
Thanks. Please take a look into libfalltergeist source code. Maybe you'll find a way to fix raw pointers issues there

PR #574, opened 2019-05-02
take a stab at adding a roadmap

sandsmark:
this is basically mostly to get some feedback on what I should start on if I want to contribute. :-)

I based it on the roadmap issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,3 +32,14 @@ or mount CD-ROM with original game, then run
 ```
 ./falltergeist
 ```
+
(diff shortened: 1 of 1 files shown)


alexeevdv:
Seems ok. Not sure if we need it in readme :)

sandsmark:
I put it there because people usually end up just reading the readme, and then forget to update it. :-P

but I could put it in a separate file, with a link from the readme?

PR #578, opened 2019-09-22
Basic barter system

JanSimek:
**What works:** buying and selling items to vendors
**What does not work:** items have fixed prices and modifiers are not yet taken into account (barter skill, Master Trader, Reputation, etc.(?))

Based on [Zervox's](https://github.com/Zervox/falltergeist/commit/32e1bc335ed4c43ae7d5b6e3ed41e5cc83663239) patch.


alexeevdv:
Can you please update it to develop branch?

PR #579, opened 2019-09-22
Fix and update Travis build

JanSimek:


alexeevdv:
Thank you for contribution

PR #581, opened 2019-09-28
Load critter flags from PRO file

JanSimek:
I saw it done like this in `Game::Object#setFlags`

Otherwise, I would have added an enum class for the flags and a method `CritterObject::hasFlag()`:

```c++
enum class CritterFlags : uint_32_t {
    Barter          = 0x00000002; // can trade with
    Steal           = 0x00000020; // cannot steal from)
    Drop            = 0x00000040; // doesn't drop items)
    Limbs           = 0x00000080; // can not lose limbs)
    Ages            = 0x00000100; // dead body does not disappear)
    Heal            = 0x00000200; // damage is not cured with time)
    Invulnerable    = 0x00000400; // cannot be hurt)
    Flatten         = 0x00000800; // leaves no dead body)
    Special         = 0x00001000; // there is a special type of death)
    Range           = 0x00002000; // melee attack is possible at a distance)
    Knock           = 0x00004000; // cannot be knocked down)
};
```
Let me know if you want it done this way or completely differently. Also, I changed the flag names from the original ones described in the [FalloutMods wiki](https://falloutmods.fandom.com/wiki/PRO_File_Format).


src/Game/CritterObject.h
```diff
@@ -178,6 +213,18 @@ namespace Falltergeist
                 int _sequence = 0;
                 int _criticalChance = 0;
 
+                bool _canTradeWith;
```
JanSimek: Perhaps _isMerchant sounds better?

alexeevdv:
Looks great. Can you also update changelog please?

FakelsHub:
Это должно быть одним полем, а не много под каждый флаг.
Придерживайтесть структуры данных движка Fallout, иначе код потом повторно придется переписывать.
Хотя я и так вижу, что многое нужно уже переписывать, (касается полей и данных объектов).

src/Game/CritterObject.h
FakelsHub: No. Just _canTrade

src/Game/CritterObject.cpp
```diff
@@ -290,6 +290,21 @@ namespace Falltergeist
             _traitsTagged.at((unsigned)num) = value;
         }
 
+        void CritterObject::setCritterFlags(unsigned int flags)
```
FakelsHub: this is cumbersome code.

JanSimek: It's done this way in `Object::setFlags` as well. I wouldn't mind rewriting the code to keep flags in a single variable, but my main intention is to keep the style the way @alexeevdv wants it.

Refactoring code like this in the future would not be a big issue. The main problem we currently face IMHO is how to get rid of all those raw pointers in the engine.

alexeevdv:
I think it is ok for now as it is. Thanks!

PR #583, opened 2019-09-28
Fix linking release builds on Windows

JanSimek:


JanSimek:
- fixed linking Release, RelWithDebInfo and MinSizeRel builds
- prevent immediate closing of console window after a crash, so the user has a chance to see what's wrong
- group source files into Headers and Sources in Visual Studio. QtCreator does this automatically and I don't know about other IDEs, so I only enabled it for MSVC

msvc-tree

alexeevdv:
Can you please also update changelog?

JanSimek:
Done

alexeevdv:
Thanks!

PR #586, opened 2019-10-03
Lockpicking & Pathing

AdamFx990:
I did some of these commits a few months ago. I think these are all the changes:

- Players run to the thing they're trying to interact with

- Doors can't be opened if they're locked

- Some optimisations to hexagon code

- Other miscellaneous optimisations & refactoring


alexeevdv:
Thanks! Can you please also update changelog file?

src/Game/DoorSceneryObject.cpp
```diff
@@ -36,89 +36,101 @@
...
+}
+
+bool DoorSceneryObject::opened() const
+{
+    return _opened;
+}
+
+void DoorSceneryObject::setOpened(bool value) {
+    // Don't change if door is locked.
+    if (_locked == false)
```
Leinnan: Why don't use `(!_locked)` instead

src/PathFinding/Hexagon.cpp
```diff
@@ -71,68 +60,14 @@ void Hexagon::setPosition(const Point& pos)
...
-void Hexagon::setHeuristic(unsigned int value)
-{
-    _heuristic = value;
-}
-
 bool Hexagon::canWalkThru()
 {
-    bool canWalkThru = true;
+    // Search hex for any blocking objects...
     for (auto object : _objects)
```
Leinnan: Maybe `(const auto& object : _objects)`?

alexeevdv:
Thanks!

PR #590, opened 2019-10-12
Change data path

JanSimek:
Load unpacked data files from the `data/` subdirectory like in the original game.

Additionally, unix-specific data path pointing to the parent directory of the executable was removed - better make it consistent on all platforms and copy the data directory to the build directory instead.

Tested on linux and Windows. BTW the path separator on Windows should probably be a backslash "\" instead of a forward slash "/", but it seems to make no difference on Windows 10, so I didn't change it.


alexeevdv:
Thanks. Can you please update changelog?

JanSimek:
Damn. I always forget.

PR #593, opened 2019-10-16
Fix crash when erasing timer on newr1 map

JanSimek:
If you load newr1.map and wait for a minute the game will crash when erasing timer from std::list. 

I have _no idea why_ but replacing `std::list` with `std::vector` fixes the issue and the game does not crash anymore. It seems to be faster too (3ms vs 11ms loop processing on average).

**src/State/Location.cpp**
```c++
// timers processing
void Location::processTimers()
{
    _actionCursorTimer.think();
    _ambientSfxTimer.think();

    for (auto it = _timerEvents.begin(); it != _timerEvents.end();) {
        it->timer.think();
        if (!it->timer.enabled()) {
            it = _timerEvents.erase(it); // <- CRASH
        } else ++it;
    }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ Under development
 =======================
 - [bugfix] Opcode810D (obj_carrying_pid_obj) checks containers as well (JanSimek)
 - [bugfix] Fix linking release builds on Windows and prevent closing console after a crash (JanSimek)
+- [bugfix] Fix crash when erasing timer on newr1 map
 - [feature] Search data files in parent dir too (alexeevdv)
 - [feature] Basic barter system (Zervox, JanSimek)
 - [feature] Fix and update Travis build (JanSimek)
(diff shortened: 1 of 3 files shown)


FakelsHub:
I think the problem lies elsewhere, and not misrepresented that you have changed container to another.

FakelsHub:
> It seems to be faster too (3ms vs 11ms loop processing on average).

why do you think that with the vector it will be faster to work in the real game?
for list `remove_if` more effective than for vector (it is not necessary to move the items to be removed).

JanSimek:
I am not sure if it will be faster in the long run. I just measured the loop execution time on newr1 map and it was 8 ms faster on average when vector was used.

As I said, I have no idea why switching from list to vector fixes the crash, but I tested it several times and it really does. It would be great if someone figured out why it is so or if there is an issue elsewhere - the crash has been happening since I first compiled this engine two years ago.

PR #594, opened 2019-10-19
World map

twboc:


JanSimek:
Btw, README should mention that the **sound** directory must be copied from the Fallout 2 installation directory as well as master.dat and critter.dat. I couldn't figure out why `20car.acm` and `23world.acm` weren't being loaded from master.dat

src/State/WorldMap/WorldMap.h
```diff
@@ -0,0 +1,157 @@
...
+                // default size data
+                unsigned int mapWidth = 450;   // fallout 2 map screen width
+                unsigned int mapHeight = 442;  // fallout 2 map screen height
+                unsigned int mapMinX;      // start X point of map screen
+                unsigned int mapMinY;      // start Y point of map screen
+
+                const std::string sidePanelSprite = "art/intrface/wminfce2.frm";
+                const std::string fullPanelSprite = "art/intrface/wmapbox.frm";
+
+                std::vector<std::string> _worldMapTiles = {
```
JanSimek, line 133: Perhaps you could use `std::array` here instead? Someone suggested that in one of my previous commits, so I am passing it on :smile:

FakelsHub: You don't need any arrays here, it should be read from the intrface.lst file
_Do you specifically write a non-compatible engine?_

src/Audio/Mixer.cpp
```diff
@@ -79,6 +79,7 @@ namespace Falltergeist
         void Mixer::stopMusic()
         {
             Mix_HookMusic(NULL, NULL);
+            return;
```
FakelsHub, line 82: Write a comment on what return is used for here.

src/State/WorldMap/WorldMapEvents.cpp
```diff
@@ -0,0 +1,88 @@
...
+            auto game = Game::getInstance();
+
+            if (auto mouseEvent = dynamic_cast<Event::Mouse*>(event))
+            {
+                auto mouse = game->mouse();
+
+                // Left button down
+                if (mouseEvent->name() == "mousedown" && mouseEvent->leftButton())
+                {
+                    int mouseX = mouse->x();
```
Bedzior, line 19: `mouseX` and `mouseY` are only used as `unsigned int` below anyway, maybe statically cast them here already?

667bdrm:
It could be great if broken mouse click handler could be restored. It will be useful for later functionality implementation.

PR #601, opened 2019-10-29
Extract UI generation logic from game objects

alexeevdv:


src/Graphics/CritterAnimationFactory.cpp
```diff
@@ -0,0 +1,53 @@
+#include "../Graphics/CritterAnimationFactory.h"
+#include "../Helpers/CritterAnimationHelper.h"
+#include "../Game/Defines.h"
+
+namespace Falltergeist
+{
+    namespace Graphics
+    {
+        std::unique_ptr<UI::Animation> CritterAnimationFactory::buildActionAnimation(uint32_t armorFID, uint32_t weaponId, const std::string &action, Game::Orientation orientation)
```
JanSimek: Would be cool if someone could create a `buildRotatingAnimation` function here for the player inventory window. I tried to simulate it by changing `orientation` of the player every half a second but that does not produce the same effect as in the original engine.

Issue #580, opened 2019-09-24
Barter formula


JanSimek:
Implement formula for calculating barter prices to match the vanilla engine.

**NovaRain** came up with the [following pseudo-code](https://www.nma-fallout.com/threads/fallout-and-fallout-2-barter-formula.217810/#post-4329046) for FO2:

```c++
Master_Trader = 0.0;
if (player_has_master_trader) Master_Trader = 25.0;

barter_mod = script_modifier + reaction_modifier; // reaction_modifier is -15/0/25 for good/neutral/bad.
barter_percent = (barter_mod + 100.0 - Master_Trader) * 0.01;

if (barter_percent < 0.0) barter_percent = 0.0099999998;

barter_ratio = [(trader_barter_skill + 160.0) / (highest_barter_skill_in_party + 160.0)] * 2.0;

total_price = caps_total + item_total_cost_without_caps * barter_ratio * barter_percent;
```

FakelsHub:
There seems to be a mistake.
![2019-10-24_05-10-42](https://user-images.githubusercontent.com/21695363/67447533-ddab0280-f61c-11e9-850f-cabacb88a012.png)

PR #603, opened 2019-10-31
Perks

JanSimek:
Prerequisities copied from https://fallout.fandom.com/wiki/Fallout_2_perks

I mostly guessed image names from SKILLDEX.LST

Refs #580

### DIFF
--- a/src/Game/DudeObject.h
+++ b/src/Game/DudeObject.h
@@ -23,7 +23,7 @@ namespace Falltergeist
         {
             public:
                 DudeObject();
-                ~DudeObject() override = default;
+                ~DudeObject() = default;
 
                 void loadFromGCDFile(Format::Gcd::File* gcd);
 
--- a/src/State/Location.cpp
+++ b/src/State/Location.cpp
@@ -474,12 +474,10 @@ namespace Falltergeist
         {
             for (auto &object: _flatObjects) {
                 object->render();
-                object->hexagon()->setInRender(object->inRender());
             }
 
             for (auto &object: _objects) {
                 object->render();
-                object->hexagon()->setInRender(object->inRender());
             }
         }
 
--- a/src/Game/AmmoItemObject.cpp
+++ b/src/Game/AmmoItemObject.cpp
@@ -1,5 +1,6 @@
 #include "../Game/AmmoItemObject.h"
 #include "../UI/TextArea.h"
+#include "../VM/Script.h"
 
 namespace Falltergeist
 {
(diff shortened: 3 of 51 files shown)


FakelsHub:
Awareness must have the ID 0

src/State/Location.cpp
```diff
@@ -346,6 +346,9 @@ namespace Falltergeist
             player->setRightHandSlot(spear);
             player->setActionAnimation("aa")->stop();
             player->setPID(0x01000001);
```
FakelsHub: player PID == 0x01000000

src/Game/DudeObject.h
```diff
@@ -92,6 +96,7 @@ namespace Falltergeist
                 int _experience = 0;
                 int _statsPoints = 0;
                 int _skillsPoints = 3;
+                std::vector<unsigned> _perks;
```
alexeevdv: We know the count for perks. Therefore it is better to use `std::array` here as it will work faster

src/State/PlayerEdit.cpp
```diff
@@ -668,5 +680,125 @@ namespace Falltergeist
             _description->setPosition(backgroundPos + Point(350, 315));
             _description->render();
         }
+
+        std::vector<PerkEligibility> PlayerEdit::getSelectablePerks() const
```
alexeevdv: `std::array` would be better probably

JanSimek:
> Awareness must have the ID 0 (start)

You are right. Other enums also start at zero. I'll fix it.

> `// JINXED - a TRAIT, but is listed in PERKS.MSG as the last perk`
> 
> This perk.

I really don't think that it is a perk. At the beginning of the game it is listed among the [traits](https://fallout.fandom.com/wiki/Jinxed) and you can find it in [TRAIT.MSG](https://fallout.fandom.com/wiki/TRAIT.MSG_(Fallout_2)). The value in PERKS.MSG is just a duplicate.

FakelsHub:
> > Awareness must have the ID 0 (start)
> 
> You are right. Other enums also start at zero. I'll fix it.
> 
> > `// JINXED - a TRAIT, but is listed in PERKS.MSG as the last perk`
> > This perk.
> 
> I really don't think that it is a perk. At the beginning of the game it is listed among the [traits](https://fallout.fandom.com/wiki/Jinxed) and you can find it in [TRAIT.MSG](https://fallout.fandom.com/wiki/TRAIT.MSG_(Fallout_2)). The value in PERKS.MSG is just a duplicate.

This perk is given from a damned dog.

JanSimek:
> This perk is given from a damned dog.

Ah, right. [Pariah dog](https://fallout.fandom.com/wiki/Pariah_dog)

PR #605, opened 2019-11-03
Fix crash when moveObjectToHexagon is called with a nullptr hexagon

JanSimek:
Happens on the Broken Hills map (broken1) when Opcode80D8 is called.

My question is why does the `Location::moveObjectToHexagon` accept a nullptr hexagon at all? Is there any reason for it or should we just check for a nullptr at the beginning and log an error if that happens?

```c++
if (!hexagon) {
        Logger::error("Location::moveObjectToHexagon") << "called with a nullptr hexagon";
        return;
}
```


FakelsHub:
fallout range the tile: 0 - 39999

Issue #608, opened 2020-02-14
Removing timer from location causes the game to hang


AdamFx990:
After selecting a character, the game hangs and the process needs to be killed manually.

EDIT: This seems to be issue on maps with rats.

### Expected Behavior

The game should work as it did before <a href="https://github.com/falltergeist/falltergeist/commit/31d3351198f1dda08bca6628aac678ffca703bc1">this commit.</a>

### Steps to Reproduce
1. Compile with g++
2. Press 'new game'
3. Select character & start game

### System
- **Falltergeist:** 31d3351198f1dda08bca6628aac678ffca703bc1
- **Operating System:** Fedora 30
- **Fallout:** Steam

### Notes

I've created a branch from this commit. I will create a pull request when I have a fix

AdamFx990:
Fix is in pull request: #609

vvbalashoff:
commit 0ceae6e5c3beb0a4da6c27b174622b45c8c83d4e

[INFO] [MESSAGE] You are once more in the village of your birth, Arroyo.
[INFO] [MESSAGE] You passed the trials of Arroyo.
[INFO] [MESSAGE] You gain 600 experience points.

game hangs, cpu usage 100%

alexeevdv:
@vvbalashoff this should be fixed in #612 
can you check if it works for you?

alexeevdv:
@AdamFx990 can you check if it fixes your issue too?

vvbalashoff:
cadd5c307c14afc42ad7978d175fe1ccc018153c fix hangs

PR #609, opened 2020-02-14
Fix for infinite loop bug

AdamFx990:
Worlds smallest pull request.

Fixes issue:  #608


FakelsHub:
Wrong fix 

**Return value**
Iterator following the last removed element.
If pos refers to the last element, then the end() iterator is returned.
If last==end() prior to removal, then the updated end() iterator is returned.
If [first, last) is an empty range, then last is returned. 

https://en.cppreference.com/w/cpp/container/list/erase

AdamFx990:
When you say to _"check [my] compiler capabilities"_: do you mean there might be a discrepancy in how this code is executed between compilers? O.o 

Are you able to replicate this issue?

FakelsHub:
> Could you please elaborate on what you mean when you say

just check "your" compiler for a possible bug
https://rextester.com/VCV91745 <- how it should work.

AdamFx990:
Thanks for the snippet. I compiled it with clang and g++ on my machine and the output was fine. 

Prior to 31d3351198f1dda08bca6628aac678ffca703bc1, the iterator behaves as intended.

alexeevdv:
Hey, thanks for the contribution. I see one major issue with you code. It adds dependency on global object (Game) to timer class. You should pass this dependency into constructor or use getters/setters to change timer object state

src/Game/Timer.h
```diff
@@ -32,6 +32,7 @@ namespace Falltergeist
                 bool _repeat = false;
                 float _interval = 0;
                 float _timeTracked = 0;
+                uint  _lastTick;
```
alexeevdv: use uint32_t please

AdamFx990:
There's been significant performance regressions on maps where _Opcode80F0_ is fired a lot.

alexeevdv:
#608 seems to be fixed

PR #617, opened 2020-04-13
Added 'x', 'y', and 'always_on_top' video options

adamkewley:
This PR adds `x`, `y`, and `always_on_top` to Falltergeist's video configuration options. 

It's a small quality-of-life patch that makes the change --> compile --> view feedback a little bit shorter because, ahead-of-time, I can configure `always_on_top = true`, `x = 1200` (on my monitor), and `y = 1200` (same) so that, whenever I run Falltergeist, the window just pops up in the bottom-right, so I can edit things in my IDE as I click around in the UI.

Again, this is nothing major and just a nice-to-have. I'm currently looking to make minor bugfixes, rather than big architectural changes, so I'm mostly speed-limited by trying something out /w a debugger, stepping through (which, because of `always_on_top`, I can visually see while still having a debugger full-screened), etc.

### DIFF
--- a/src/Game/Game.cpp
+++ b/src/Game/Game.cpp
@@ -23,6 +23,26 @@
 #include "../UI/FpsCounter.h"
 #include "../UI/TextArea.h"
 
+namespace
+{
+    Falltergeist::Graphics::RendererConfig cfg_from_settings(const Falltergeist::Settings& settings)
+    {
+        Falltergeist::Graphics::RendererConfig cfg{
+                .width = settings.screenWidth(),
+                .height = settings.screenHeight(),
+                .fullscreen = settings.fullscreen(),
+                .alwaysOnTop = settings.alwaysOnTop(),
+        };
+        if (settings.screenX() >= 0) {
+            cfg.x = settings.screenX();
+        }
+        if (settings.screenY() >= 0) {
+            cfg.y = settings.screenY();
+        }
+        return cfg;
+    }
+}
+
 namespace Falltergeist
 {
     namespace Game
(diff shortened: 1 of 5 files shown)


adamkewley:
Apologies: forgot to branch this onto an appropriate feature branch (e.g. `feature-add-alwaysontop`) and this PR is adsorbing some other refactors I have been working on (which I'll PR after this, if you want them).

src/Game/Game.cpp
```diff
@@ -23,6 +23,26 @@
 #include "../UI/FpsCounter.h"
 #include "../UI/TextArea.h"
 
+namespace
+{
+    Falltergeist::Graphics::RendererConfig cfg_from_settings(const Falltergeist::Settings& settings)
```
alexeevdv, line 28: Lets use camel case for method names

src/Game/Game.cpp
```diff
@@ -23,6 +23,26 @@
...
+    {
+        Falltergeist::Graphics::RendererConfig cfg{
```
alexeevdv, line 30: Can you please use full name? Like `config` instead of `cfg`. Otherwise it affects readability of the code

src/Settings.h
```diff
@@ -23,9 +23,13 @@ namespace Falltergeist
             bool save();
             bool load();
 
-            unsigned int screenWidth() const;
+            int screenWidth() const;
```
alexeevdv, line 26: As far as i concerned screen width can not be negative :)

adamkewley:
Thanks for the comments @alexeevdv . I'll look into refactoring them. 

As for the width being non-negative, would `0` be an acceptable value for "not provided" or "default"?

adamkewley:
@alexeevdv this might be closer to what you want.

If you think this looks good, I can squeeze this changeset down into one commit and force-push it to this PR so that the review changes etc. don't pollute the logs

adamkewley:
Switched `RendererConfig` to be taken as a `const &` and fixed up some other signed-to-unsigned changes

alexeevdv:
Merged. Thanks. I will format a code a little bit and then it is going to be merged to develop branch

adamkewley:
Ah great, thanks for your feedback @alexeevdv

PR #618, opened 2020-04-18
WIP: refactoring State to use smart pointers

adamkewley:
This is a work-in-progress refactor.

The `State` base classed used in *many* places in Falltergeist currently uses bare pointers. Bare pointers aren't, by themselves, the worst thing in the world, but they're potentially quite nasty in this particular part of Falltergeist because it's unclear what the order of destruction is when (e.g.) states are popped but other parts of components depend on pointers to now-stale elements, etc. etc.

Maybe it's because COVID is making me go crazy or maybe, deep down, I just prefer refactoring to adding new bugs, but this PR covers refactoring the `State` to use shared pointers. 

I will try to limit this work to only adding shared-pointer support, but if I see any benign-looking refactors as I move forward I will also do them (provided their diffs make sense in the greater context of the work).

### DIFF
--- a/main.cpp
+++ b/main.cpp
@@ -1,41 +1,22 @@
-#include <memory>
-#include "src/Exception.h"
 #include "src/Game/Game.h"
 #include "src/Logger.h"
 #include "src/Settings.h"
 #include "src/State/Start.h"
 #include "src/UI/ResourceManager.h"
 
+#include <memory>
+
 using namespace Falltergeist;
 
 int main(int argc, char* argv[])
 {
-    try
-    {
-        auto uiResourceManager = std::make_shared<UI::ResourceManager>();
+    auto uiResourceManager = std::make_shared<UI::ResourceManager>();
 
-        auto game = Game::Game::getInstance();
-        game->setUIResourceManager(uiResourceManager);
-        game->init(std::unique_ptr<Settings>(new Settings()));
-        game->setState(new State::Start(uiResourceManager));
-        game->run();
-        game->shutdown();
-        return 0;
-    }
-    catch(const Exception &e)
-    {
-        Logger::critical() << e.what() << std::endl;
-
-#if defined(_WIN32) || defined(WIN32)
-        system("PAUSE");
-#endif
-    }
-    /*
-    catch (const std::exception &e)
-    {
-        Logger::critical() << e.what() << std::endl;
-    }
-    */
-    return 1;
+    auto game = Game::Game::getInstance();
+    game->setUIResourceManager(uiResourceManager);
+    game->init(std::make_unique<Settings>());
+    game->setState(std::make_unique<State::Start>(uiResourceManager));
+    game->run();
+    game->shutdown();
+    return 0;
 }
-
--- a/src/Event/EventTarget.h
+++ b/src/Event/EventTarget.h
@@ -1,7 +1,7 @@
 #pragma once
 
-#include <memory>
 #include "../Event/Handler.h"
+#include <memory>
 
 namespace Falltergeist
 {
--- a/src/Format/Msg/File.cpp
+++ b/src/Format/Msg/File.cpp
@@ -89,7 +89,18 @@ namespace Falltergeist
                         return &message;
                     }
                 }
-                throw Exception("File::message() - number is out of range: " + std::to_string(number));
+                throw std::runtime_error("File::message() - number is out of range: " + std::to_string(number));
+            }
+
+            bool File::hasMessage(unsigned int number) {
+                for (auto& message : _messages)
+                {
+                    if (message.number() == number)
+                    {
+                        return true;
+                    }
+                }
+                return false;
             }
         }
     }
--- a/src/Format/Msg/File.h
+++ b/src/Format/Msg/File.h
@@ -20,6 +20,7 @@ namespace Falltergeist
                 public:
                     File(Dat::Stream&& stream);
 
+                    bool hasMessage(unsigned int number);
                     Message* message(unsigned int number);
 
                 private:
--- a/src/Game/ContainerItemObject.cpp
+++ b/src/Game/ContainerItemObject.cpp
@@ -20,9 +20,9 @@ namespace Falltergeist
 
         void ContainerItemObject::use_p_proc(CritterObject* usedBy)
         {
-            auto state = new State::Container(std::make_shared<UI::ResourceManager>());
+            auto state = std::make_unique<State::Container>(std::make_shared<UI::ResourceManager>());
             state->setObject(this);
-            Game::getInstance()->pushState(state);
+            Game::getInstance()->pushState(std::move(state));
         }
 
         void ContainerItemObject::setLocked(bool locked)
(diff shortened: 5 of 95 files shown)


adamkewley:
Ok, so this cute little change, of making the base `State` class use smart pointers, seems to hit basically every component in `State/`.

I'm experimenting with methods that use template forwarding to add UI components to `State` more straightfowardly. One thing that pops out while I'm doing this is that there's *a lot* of code for labelling and getting states from the underlying `std::map`, and that `State`'s main utility for having an `addUI` method is to abstract away a for loop that just renders every subcomponent.

It might actually be an idea to make `State` not have an implementation of `render` (that for loop) because it's only a few lines to implement it - it'd be better if each class that derives `State` just supplies their own `render` function because it means that they can directly initialize + hold all UI subcomponents in their own implementation, rather than relying on the black-box of calling `addUI`.

JanSimek:
Finally! Someone experienced decided to tackle this difficult refactoring task. 

Just one question. What's the purpose of doing this? Couldn't you just keep the string parameter as a constant reference?

```diff
- TextArea::TextArea(const std::string& text, int x, int y) : TextArea(text, Point(x, y))
+ TextArea::TextArea(std::string text, int x, int y) : TextArea(std::move(text), Point(x, y))
```

adamkewley:
@JanSimek the `TextArea` owns a copy of the `std::string`. 

In the first one (a constant ref), the reference is given to the ctor, which forwards it to the overload (another ctor), which copies the reference into a member variable in `TextArea`, so you end up copying the provided string into the `TextArea`.

In the second one (a value), the value is given to the ctor--a value which, crucially, *may* have been constructed by its move-constructor--the ctor then `std::move`s the value into the other ctor, which also takes a value (this value is *definitely* move-constructed), and the other ctor `std::moves` the string into its final member-variable location.

The second one seems like it's more work on the surface, but keep in mind a few details:

- The first one *always* copies the string, so you will *always* end up with two copies of the string, even if the string you provide it is ultimately an rvalue (unless LTOd optimization figured out is unnecessary, but you need to do an `-flto` optimization for that, and it doesn't always work)
- The second one, at worst, copies the string (same as first) *and* does all the moving around so, worst-case, it's as bad as the first + 2 move constructions. Assuming move construction is cheap, that worst-case situation isn't likely to be bad
- The seconds one's constructors are all in one `.cpp` file, the same compilation unit, so it's very likely that the compiler will optimize the constructor forwarding and assignment into a single assignment step.

So in the refactored version, if you build up a string on the caller's side then `std::move` it into the constructor, or just pass some other rvalue (func return, whatever) into the refactored version, you'll only create one string, which will probably be assigned to the members directly (post-optimization).

adamkewley:
w.r.t. some of these refactors, I'm usually switching refs for value types if it feels like there's no huge benefit to using a reference type. Similar logic applies to changing many of the `Point` etc. functions to value types: ultimately, passing a 2*`int` point by value isn't going to be much more expensive than passing the reference.

I wouldn't do this for something that was expensive to construct (e.g. a string): for those, I'll generally check if the receiver of the argument is ultimately going to copy the reference anyway and just refactor the method to take a value (e.g. `TextArea`)

adamkewley:
After reviewing how many callsites I'd need to change for this refactor, I might need to dump a lot of this work and, instead, figure out how to gradually introduce a cleaner abstraction that "fits" with the rest of `Falltergeist` but is amenable to gradually introducing these refactors one-by-one (e.g. a `PureState` abstract base that is used by the engine, keep `State` as-is, gradually move components over to the new `PureState`)

At the moment, it looks like this refactor's going to be >2-3 kLOC before the thing even compiles, which requires more discipline than I have (writing 2-3 kLOC that compiles is the easy part, avoiding introducing regressions is where this will be difficult)

adamkewley:
This initial refactor's working, but I can't create a game. It looks like the `DudeObject` created in the `NewGame` state hasn't had `setFID` called on it, which means it has no `ui`, which segfaults when starting a new game.

adamkewley:
Ok, so after *a lot* (hours) of debugging into why creating a `New Game` in the Falltergeist can segfault if I first load directly into a game (via config option) followed by exiting and creating a new one, it turns out that there's a cute `static` variable that gets set when the player loads for the first time which, via several layers of indirection, means that an innocent-looking initialization function doesn't call another innocent-looking "set animation function", which means that the resulting `DudeObject` ends up with no `_ui` variable set, which results in a null dereference.

Although smart pointers might help with the leaks, it's clear from this debugging wave that the current codebase has *a lot* of bizzare indirect methods that need to be called on some objects in order to initialize them into a valid state (no, not just `initialize` - this error is because `_generateUI` isn't called, which is called via several other functions).

I think, if I'm going to refactor this thing into shape, it's going to require some pretty meaty changes to how the various objects in the game engine are initialized (crucially, via their constructors, with exceptions on failure, etc.). The current design is just an open invite for a dev to forget to call a member at the right time and *boom* - bye bye afternoon.

adamkewley:
Alright, so now the core state-handling `Game` API has been refactored to use smart pointers, rather than raw pointers, so we can get a handle on state lifetime.

Turns out, almost all states are owned entirely by the `Game` object, apart from ~4 cases, in critter dialog, where the dialog classes seem to depend on mutating other dialog states and non-active dialog states, before they are sent to `Game`: if they can be cleaned up, then the `Game` object can take exclusive ownership of the game state and, so, it'll be much more obvious what the lifetime of the game state actually is (essentially, between `push` and `pop`).

alexeevdv:
@adamkewley Thanks for your efforts. I will take a closer look for what you are proposing soon. I had same plan about smart pointers but i also want to get rid of UI elements and replace them with Entities. Game objects and UI elements should be both entities. This will simplify states a lot and will reduce coupling between the classes

adamkewley:
@alexeevdv thanks for looking into this. I'm currently just trying to remove all uses of *raw* pointers, replacing them with shared_ptr for now: seems to be compiling + working. 

The next step, I think, is to discriminate between adding shared vs. uniquely owned UI elements with different method names (e.g. `addUI`, `addUIShared`) because I think that most uses of `addUI` *probably* just want a temporary reference to a uniquely-owned UI element (to initialize it etc.). However, there's a few edge-cases where the pointer is saved, creating "shared ownership", which I've resolved by using shared pointers, but I think that these are limited use-cases (e.g. when a state wants to keep a copy of the shared pointer around to use it at runtime).

Ultimately, going through the `State` stuff, I think loads of the UI elements/states would be better-served by just inheriting from a few lighter-weight interfaces and having them render their own substates. Instead of each state calling `addUI`, just assign to a member variable and implement a `render` that calls on each sub-state. This will eliminate the need for `getUI`, which is how many states get the substate back out of the base class at runtime.

This is a pretty big diff though @alexeevdv - please just treat it as an exploration exercise: I can maybe cut this down into some smaller PRs that I can submit separately, because there's a lot to review in this - I'm bored and treating this as a bit of fun, so I won't take it personally if you reject this PR.

adamkewley:
Alright, so now `State` doesn't handle raw pointers, only smart ones.

PR #619, opened 2020-04-26
Fix error during glew loading under Wayland

ids1024:
Solution taken from Supertux: https://github.com/SuperTux/supertux/pull/808


alexeevdv:
Thanks for contribution

PR #620, opened 2020-04-27
Updated project for up-to-date Windows building (conanfile, gitignore, and README)

s7jones:
As mentioned in the title - these are the changes I needed to do to get this project to compile on Windows.

I cannot run the game, due to a problem with SDL_image (https://github.com/bincrafters/community/issues/981). Hopefully we can fix that soon.

Thanks.

### DIFF
--- a/conanfile.txt
+++ b/conanfile.txt
@@ -5,6 +5,7 @@ sdl2_image/2.0.5@bincrafters/stable
 sdl2_mixer/2.0.4@bincrafters/stable
 glew/2.1.0@bincrafters/stable
 glm/0.9.9.5@g-truc/stable
+libiconv/1.16@_/_
 
 [generators]
 cmake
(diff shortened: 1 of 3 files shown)


conanfile.txt
```diff
@@ -5,6 +5,7 @@ sdl2_image/2.0.5@bincrafters/stable
 sdl2_mixer/2.0.4@bincrafters/stable
 glew/2.1.0@bincrafters/stable
 glm/0.9.9.5@g-truc/stable
+libiconv/1.16@_/_
```
Croydon, line 8:
```suggestion
libiconv/1.16
```

`@_/_` are just used internal and mean no username and no channel. Within `conanfile.txt` and `conanfile.py` you can and should omit this. On the command line you need to add only a `@` at the end for backwards compatibilty like `libiconv/1.16@` (this will change with Conan 2.0)

Croydon: Packages from Conan Center index do not have that https://github.com/conan-io/conan-center-index

username is unnecessary as it is maintained by the entire community and the channel is unnecessary as we don't maintain `testing` packages or similar

s7jones: OK thanks for letting me know.

alexeevdv:
Thanks

PR #629, opened 2020-06-05
Extend README with mention of SDL conan remote

Bedzior:
On the first run I realized some conan packages were missing. Luckily I found them in bincrafters' repository and would like to share this in the README.


alexeevdv:
Thanks!

PR #638, opened 2020-07-18
Added 811A, 80E6, 80EE opcode handlers

667bdrm:
This fixing ending game on walking on traps at raiders1.map, rndbess.map, bhrnddst.map


alexeevdv:
Thanks!

PR #643, opened 2020-07-21
Fix locked doors

667bdrm:
Checking door lock on use_p_proc, playing locked sound if door locke and don't start opening animation for locked door.


alexeevdv:
Thanks

PR #648, opened 2020-07-25
Visual Studio Code / VSCodium build and debug tasks for Windows

667bdrm:
Added tasks for Visual Studio Code / VSCodium for dependencies install and build automation in Visual Studio / Microsoft C++ Build Tools. Based on AppVeyor build configuration. PowerShell script could be also used standalone for sln generation.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -4,8 +4,9 @@
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
     "configurations": [
+
         {
-            "name": "(gdb) Launch",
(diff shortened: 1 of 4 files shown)


alexeevdv:
I have no idea what is going on here but since it is in `.vscode` folder i dont care :) thanks

667bdrm:
There are makefile tasks like in AppVeyor script. Used PowerShell because it included into many modern windows and it has built-in download command. So basically to start builing needs to have only  VS Code and Microsoft C++ build tools (or Visual Studio) - the dependencies will be downloaded and installed into build dir. So I hope this help someone write a good feature or fix big bug :)

PR #650, opened 2020-07-28
Add very simple scripts to build the project in Docker

dreamer-dead:
Hi folks.

What do you think about ability to build the project in a docker container?
I'd like to play a bit with the code, but building on real machine may be inconvenient.
So here are few simple scripts to make local docker builds possible.
Main intent is to build current working dir in Linux environment, maybe with different compilers or options.
I can push resulting docker image to a public registry if needed.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,16 @@
+FROM ubuntu:18.04
+
+RUN apt-get update --fix-missing && \
+  apt-get install -y \
+  build-essential \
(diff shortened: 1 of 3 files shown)


alexeevdv:
You need to install libs on host machine to be able to run it anyway

alexeevdv:
You can build binaries in Docker or with any other tool on local. But i see no reason to add this in project repo. We have github actions for this particular purpose

PR #659, opened 2021-12-21
Inline trim functions

eirnym:


eirnym:
Which cpp version do you use? 

Mine is following:
```
Apple clang version 12.0.0 (clang-1200.0.32.29)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

alexeevdv:
It would be great to make them generic then. Can you please add possibility to trim any symbol and not just space?

eirnym:
> It would be great to make them generic then
Yes, I think it's possible. 

@alexeevdv could you please check why it doesn't want to compile on Ubuntu? Should I install another compiler?

eirnym:
This project doesn't have any tests. Which test suite do you prefer? Google tests, CTests or something else?

alexeevdv:
I do not see any use of tests at this point

eirnym:
I'd love to introduce them

alexeevdv:
I would like to see tests in separate branch then. Do not mix it with any other changes please

PR #660, opened 2021-12-21
Fix incompatible cast in VertexArray

eirnym:
Cast from `unsigned int` to `void *` for small amounts of data it's ok, but it could lead to
unpredictable results in the future.


src/Graphics/VertexArray.cpp
```diff
@@ -25,7 +25,7 @@ namespace Falltergeist {
         void VertexArray::addBuffer(const std::unique_ptr<VertexBuffer>& buffer, const VertexBufferLayout &bufferLayout) {
             bind();
             buffer->bind();
-            unsigned int offset = 0;
+            void * offset = 0;
```
eirnym, line 28: alternatively it could be a `char *`, which reduce a cast at line `57`

alexeevdv:
Thanks

PR #670, opened 2022-05-18
CMake missing sdl2main lib

MrCool92:
I have noticed that sdl2main.lib was not linked with CMake.


alexeevdv:
Thanks! It will be merged with https://github.com/falltergeist/falltergeist/pull/671 as i need to fix windows build first

PR #671, opened 2022-05-20
sdl2main.lib is not linked with CMake (MrCool92)

alexeevdv:


alexeevdv:
Originated from https://github.com/falltergeist/falltergeist/pull/670

PR #674, opened 2022-05-20
Temp fix for acm sound length

MrCool92:
len_ratio is 0 making all sounds length 0


alexeevdv:
Thanks! I will try to reproduce it on Windows next week. Hope it is known SDL bug

PR #681, opened 2022-06-11
Worldmap cities

MrCool92:
- Worldmap city circles rendering
image
I still need to figure out how to render actual cities in circles...

- Fixed .fon file loading (big endian stream -> little endian stream)


alexeevdv:
Looks great. Thanks!

